PR #2, opened 2015-10-07
Enumerate alternative PCIe bus

netwerkforens:
Start enumerating at a specific bus number, this allows accessing root
complexes that do not start a busnumber 0. (e.g. Intel Corporation Xeon
E5/Core internal components and root complex on second CPU).


netwerkforens:
Rejected by comment on mailing lis

PR #5, opened 2018-05-15
Add device-tree node path to the verbose output

oohal:
Adds the path of the device-tree node of a PCI device to the lspci -v
output, like so:

0021:00:00.0 PCI bridge: IBM Device 03dc (prog-if 00 [Normal decode])
	DT Node: /sys/firmware/devicetree/base/pciex@3fffe41100000/pci@0

This is added as a generic property to struct pci_device and populated
by the sysfs backend. Other platforms may find it useful though.

Signed-off-by: Oliver O'Halloran <oohall@gmail.com>

### DIFF
--- a/lib/access.c
+++ b/lib/access.c
@@ -75,6 +75,7 @@ void pci_free_dev(struct pci_dev *d)
   pci_mfree(d->module_alias);
   pci_mfree(d->label);
   pci_mfree(d->phy_slot);
+  pci_mfree(d->dt_node);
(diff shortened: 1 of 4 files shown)


gollux:
I like adding information on DT nodes, but your implementation breaks ABI compatibility, because it introduces new public struct fields without updating symbol versions.

I chose a more general solution, could you please check that the current master branch works for you?

oohal:
That does the trick. Thanks!.

PR #6, opened 2018-07-01
lib/sysfs.c: use realpath instead of canonicalize_file_name.

maxice8:
both are the same and canonicalize_file_name is glibc specific, using
realpath allows pciutils to compile on musl libc systems like Void Linux
and Alpine Linux.

### DIFF
--- a/lib/sysfs.c
+++ b/lib/sysfs.c
@@ -133,7 +133,7 @@ sysfs_deref_link(struct pci_dev *d, char *link_name)
   sysfs_obj_name(d, "", path);
   strcat(path, rel_path);
 
-  return canonicalize_file_name(path);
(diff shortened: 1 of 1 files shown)


maxice8:
build log of what happens on x86_64-musl https://travis-ci.org/void-linux/void-packages/jobs/398717651

dtzWill:
In case it helps, from `(man 3 canonicalize_file_name`](https://linux.die.net/man/3/canonicalize_file_name):

```
The call canonicalize_file_name(path) is equivalent to the call:

realpath(path, NULL);
```

In case there were any doubts about this change, it just so happens to be documented to be equivalent :).

gollux:
I applied a similar fix to master and released version 3.6.1.

PR #1, opened 2015-10-05
Enumerate all undiscovered PCIe busses

netwerkforens:
Multiple root complexes have a different start busnumber.
Poll all undiscovered busses for a root complex

Signed-off-by: Ruud <netwerkforens@gmail.com>

PR #7, opened 2018-07-18
Enable setpci to target n-th capability of id

JohnAZoidberg:
Because a capability can exist multiple times with the same id
there needs to be a way to target a specific one. Instead of
the current behaviour which always targets the first one.
Now you can optionally add `@number` (e.g `@1`) after the width to
choose which one to target.

Please let me know what you think :)

### DIFF
--- a/lib/caps.c
+++ b/lib/caps.c
@@ -104,11 +104,21 @@ pci_free_caps(struct pci_dev *d)
 struct pci_cap *
 pci_find_cap(struct pci_dev *d, unsigned int id, unsigned int type)
 {
+  return pci_find_cap_nr(d, id, type, 0);
+}
+
+struct pci_cap *
+pci_find_cap_nr(struct pci_dev *d, unsigned int id, unsigned int type, unsigned int cap_number)
+{
   struct pci_cap *c;
+  unsigned int num = 0;
 
   pci_fill_info_v35(d, ((type == PCI_CAP_NORMAL) ? PCI_FILL_CAPS : PCI_FILL_EXT_CAPS));
   for (c=d->first_cap; c; c=c->next)
-    if (c->type == type && c->id == id)
-      return c;
+     if (c->type == type && c->id == id)
+       if (num == cap_number)
+          return c;
+       else
+          num++;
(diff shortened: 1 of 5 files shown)


gollux:
Generally speaking, this is a nice feature. However, you need to implement it in a way which does not break API nor ABI of the library. Instead of adding a new parameter to an existing function, create a new function.

JohnAZoidberg:
I'm no expert on ABI compatibility but this should do it, right?
Is it correct that I added the new function in the `LIBPCI_3.6` section of `lib/libpci.ver`?

Because indexing starts at 0 when it doesn't find the capability of number 1 it says:
`setpci: 0000:02:00.0: 1st Capability 0013 not found`
I think that might be confusing. Because it's not the first but the one at index 1.

gollux:
> I'm no expert on ABI compatibility but this should do it, right?

Yes.

> Is it correct that I added the new function in the `LIBPCI_3.6` section of `lib/libpci.ver`?

Please add a new section LIBPCI_3.7.

> Because indexing starts at 0 when it doesn't find the capability of number 1 it says:
> `setpci: 0000:02:00.0: 1st Capability 0013 not found`
> I think that might be confusing. Because it's not the first but the one at index 1.

Yes, it is somewhat confusing. What about "Instance #1 of capability 0013 not
found"?

JohnAZoidberg:
I thought a while about how to best word the error.
I think your suggestions goes in the right direction. Additionally it should show how many capabilities there actually are to avoid confusion. Because even with `#1` someone might think that it's the first one.

What I suggest is:
If there are no capabilities with that id:
`setpci: 0000:02:00.0: Instance #1 of Capability 0001 not found - there are no capabilities with that id`

When the user wants `@1`(the second one) but there is only 1:
`setpci: 0000:02:00.0: Instance #1 of Capability 0001 not found - there is only 1 capability with that id`

If we tell the user how many they are they will know that's not how you address the first one.

gollux:
I squash-merged this branch and cleaned up pci_find_cap_nr(). Please check.

PR #9, opened 2018-08-13
Fix solaris build

andy-js:
Defining BYTE_ORDER on Solaris breaks the arpa headers.

### DIFF
--- a/lib/names-net.c
+++ b/lib/names-net.c
@@ -15,6 +15,14 @@
 
 #ifdef PCI_USE_DNS
 
+/*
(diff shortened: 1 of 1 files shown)


gollux:
#undef'ing it on all platforms looks dangerous. Please do it only on Solaris.

ralic:
It would be better to open a branch for Solaris and keeps this on that branch only.

gollux:
No, it wouldn't. Please fix the problem properly.

andy-js:
Done.

gollux:
Merged, thanks.

PR #10, opened 2018-08-13
Pass PREFIX to configure

andy-js:
Pass PREFIX to configure so that libraries get installed into the right directory.


gollux:
I do not see any sense in this: our configure script takes no arguments.

andy-js:
It takes all the standard autoconf arguments.  Run it with "--help" and you'll see what I mean.

andy-js:
Never mind.  I take that back.  I'll close this.

PR #13, opened 2018-09-18
lspci: allow -s option to be used with -t to show a subtree

gera-k:


gollux:
I merged the first change.

I don't like the second one as it introduces a lot of code duplication and the feature itself is of dubious interest.

PR #14, opened 2018-09-19
Updated pci.ids and makefile

ralic:


gollux:
This cannot be merged: it contains binary DLLs (worse, they have no license declared), it makes random changes to pci.ids (please submit them via pci-ids.ucw.cz web interface), and it breaks build for everybody else.

ralic:
Would it be better to open another branch for windows version ?

gollux:
No, especially because files with no clear license do not belong to the repository at all.

If you want to fix Windows builds, please do it properly.

PR #16, opened 2018-10-03
Use HTTPS when possible.

MilanKral:


gollux:
Most changes have been merged.

It is pointless to patch pci.ids, since it is an auto-generated file.

PR #23, opened 2019-01-25
 Set PCI_HAVE_64BIT_ADDRESS for NetBSD.

msaitoh:
When I set "Over 4G decoding" in a BIOS setting and boot NetBSD machine,
I've noticed that lspci reports:
    pcilib: 0000:07:00.0 64-bit device address ignored.
and
    Region 0: Memory at <ignored> (64-bit, prefetchable)

After applying this change:
    Region 0: Memory at 7d7f600000 (64-bit, prefetchable)


gollux:
Thanks, applied.

PR #25, opened 2019-02-03
lspci: Fix extra newline if L1.2 is not supported.

vinsonlee:
Fixes: fb17077dc378 ("Cleaned up the previous patch")
Signed-off-by: Vinson Lee <vlee@freedesktop.org>


gollux:
Applied, thanks.

PR #29, opened 2019-10-03
libpci.pc.in: Use prefixes instead of hardcoding

neheb:
This makes it more cross compile friendly.

Still need a solution for libdir since it can be either lib or lib64 (and maybe lib32).


gollux:
Could you please explain why is this change useful?

I suspect that while it makes some use cases more straightforward, it could make others worse.
In particular, when somebody wants to install pci.ids to a different directory, it sufficed to change IDSDIR, but after your change, it is necessary to edit libpci.pc in addition to that.

neheb:
Meh i'll keep it local.

PR #30, opened 2019-12-05
lspci: Avoid printing null character in VPD fields

ptalbert:
An 'Unknown' VPD field with 0x0 id is translated to an ascii NULL character
by printf() %c format specifier. This makes for some ugly output and tools
such as gawk will throw a warning when parsing it.

Avoid this by using the %s format specifier with a length limit of 1.

Signed-off-by: Patrick Talbert <ptalbert@redhat.com>

### DIFF
--- a/ls-vpd.c
+++ b/ls-vpd.c
@@ -178,7 +178,7 @@ cap_vpd(struct device *d)
 	      if (!read_vpd(d, res_addr + part_pos, buf, read_len, &csum))
 		break;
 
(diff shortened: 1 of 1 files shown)


gollux:
Null characters definitely shouldn't make it into the output, but I suppose that if 0x00 can appear in field IDs, so could other control characters. Could you please modify the patch to escape all characters outside 0x20-0x7e?

gollux:
I implemented proper escaping. In fact, I just re-used what we already do for VPD items containing strings. Committed to master.

PR #32, opened 2019-12-27
New access method: Hurd via RPCs

jlledom:
The Hurd now uses a PCI arbiter for coordinating access and granting permissions to the PCI bus. Clients cannot use IO ports for this purpose anymore.

This patch adds a new module which uses the proper interface via RPCs to access the bus through the arbiter.


gollux:
Thanks for the new back-end. Could you please clean up the source to bring it closer to the coding style of the other modules? In particular, function calls should not contain spaces before the opening parenthesis.

Also, I would prefer keeping the I/O-port banging back-end available on Hurd as the second choice. It can be handy when debugging obscure problems, or if somebody uses an old version of Hurd which does not have the RPC interface yet.

jlledom:
Just pushed two commits with the required changes. About the coding style, the spaces before the parenthesis in function calls is the only difference I've found. Please tell me if you need something else.

lib/hurd.c
```diff
@@ -0,0 +1,381 @@
...
+static void
+hurd_cleanup(struct pci_access *a UNUSED)
+{
+}
+
+/* Each device has its own server path. Allocate space for the port. */
+static void
+hurd_init_dev(struct pci_dev *d)
+{
+  d->aux = calloc(1, sizeof(mach_port_t));
```
gollux: Please use pci_malloc()

lib/hurd.c
```diff
@@ -0,0 +1,381 @@
...
+/* Deallocate the port and free its space */
+static void
+hurd_cleanup_dev(struct pci_dev *d)
+{
+  mach_port_t device_port;
+
+  device_port = *((mach_port_t *) d->aux);
+  mach_port_deallocate(mach_task_self(), device_port);
+
+  free(d->aux);
```
gollux: Please use pci_mfree().

lib/hurd.c
```diff
@@ -0,0 +1,381 @@
...
+}
+
+/* Enumerate devices */
+static void
+hurd_scan(struct pci_access *a)
+{
+  int err;
+
+  err = enum_devices(_SERVERS_BUS_PCI, a, -1, -1, -1, -1, LEVEL_DOMAIN);
+  assert(err == 0);
```
gollux: Please do not use asserts. Report errors via the a->error callback, which can point to a custom error handler provided by the user of the library.

lib/hurd.c
```diff
@@ -0,0 +1,381 @@
...
+    {
+      d->vendor_id = pci_read_word(d, PCI_VENDOR_ID);
+      d->device_id = pci_read_word(d, PCI_DEVICE_ID);
+    }
+
+  if (flags & PCI_FILL_CLASS)
+    d->device_class = pci_read_word(d, PCI_CLASS_DEVICE);
+
+  if (flags & PCI_FILL_IRQ)
+    d->irq = pci_read_byte(d, PCI_INTERRUPT_LINE);
```
gollux: You should let pci_generic_fill_info() handle that. Generally, back-ends implement fill_info by handling specific things and then letting the generic code handle the rest.

lib/init.c
```diff
@@ -80,6 +85,7 @@ static int probe_sequence[] = {
   PCI_ACCESS_OBSD_DEVICE,
   PCI_ACCESS_DARWIN,
   PCI_ACCESS_SYLIXOS_DEVICE,
+  PCI_ACCESS_GNU,
```
gollux: Should we call this GNU if the back-end calls itself "hurd"?

lib/hurd.c
jlledom: Done in 54ca33c

lib/hurd.c
jlledom: Done in 54ca33c

lib/hurd.c
jlledom: Done in d1f0616

lib/hurd.c
jlledom: Done in 5535c72

lib/init.c
jlledom: Done in 1cedbcc

jlledom:
I've committed some changes to meet your requests

jlledom:
Sorry I've closed this by mistake :(

gollux:
Sorry, the problematic indents were caused by GitHub misdisplaying tabs. I merged all changes as a single commit. I will release version 3.6.4 in a moment. Thanks for your help!

PR #35, opened 2020-02-28
lspci: Print IOMMU groups with -v

Hello71:
PCI passthrough is increasingly popular. This makes it a little bit easier.

It doesn't seem to actually work properly though... it only works with -vv, and it doesn't work with -mmvv, and I can't figure out why.


gollux:
Overall, I like the patch. Please update the description of machine-readable output in the manual page to include the new field

It should work with `-mmvv`, too. I do not see why it shouldn't ... however, if it still doesn't, please strace it to check that it really reads the sysfs file.

Hello71:
Seems to work now... I think maybe I was running the system lspci before :s. Added man page description and rebased on master.

gollux:
Good. I just merged it, thanks for you contribution.

PR #42, opened 2020-06-06
Hurd: bug fixes and compile again

jlledom:
I've fixed a bug that produced a memory violation, and also made the module compile again since `a` was undefined in `device_port_lookup`


gollux:
Thanks, going to merge.

PR #43, opened 2020-06-19
hurd: Fix build

sthibaul:


sthibaul:
(that was actually already fixed by 053cf6c8b2acafadf828912828336d90fe9b8696)

PR #46, opened 2020-09-20
Fix typos

guillemj:
Fix several typos. Some affected the ChangeLog which I still fixed, even though some people have some aversion to touching these kinds of files.


gollux:
Thanks, applied.

PR #47, opened 2020-09-20
Build system fixes

guillemj:
Various build system fixes that have been carried over time in the Debian packaging.


gollux:
Thanks, merged.

PR #50, opened 2020-10-08
Makefile: fix for parallel builds (make -j)

zach-welch-timesys:
Without this change, the build will fail when it cannot find `libpci.pc`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ install-pcilib: lib/$(PCILIB)
 	$(DIRINSTALL) -m 755 $(DESTDIR)$(LIBDIR)
 	$(INSTALL) -c -m 644 lib/$(PCILIB) $(DESTDIR)$(LIBDIR)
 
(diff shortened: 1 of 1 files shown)


gollux:
While I admit that the problem can exist, your fix is wrong. Make does not support ordinary (non-pattern) rules with multiple targets. If you write two targets before ":", Make creates a separate copy of the rule for each target. This way, `lib/configure` could be run multiple times in parallel, which is not likely to work.

I can try to find a better fix, but I am not able to reproduce your problem.

zach-welch-timesys:
Thank you for pointing out that nuance with make. I have revised the patch to address the problem in a different manner.

The problem manifests when the `install-lib` rule looks for `lib/$(PCILIBPC)` before the `install-pcilib` rule finishes running. As the `install-pcilib` rule itself will cause that file to be built, it should be removed as a direct dependency from `install-lib`.

PR #51, opened 2020-10-30
Added five nibbles for domain in dump read

rohit-mundra:
Allow five nibbles as valid domain, when reading from a dump file with
domain


gollux:
Thanks, applied.

PR #54, opened 2020-12-12
add missing symlinks to libraries in install-lib target

lexming:
The `install-lib` target results in a broken installation because it only symlinks `libpci.so` to `libpci.so.$(ABI_VERSION)`, but `libpci.so.$(ABI_VERSION)` does not exist. This PR adds the missing symlink between `libpci.so.$(ABI_VERSION)` and the actual library file.


gollux:
Thanks!

PR #57, opened 2021-03-04
Fix the primitive system dependency discovery for Darwin systems

e820:
Checking `/usr/include` for system library headers is insufficient for Darwin as `libresolv` and `zlib` ship with the macOS SDK. This patch simply uses `xcrun` to detect the current macOS SDK root and expands it when checking for the headers.


gollux:
Applied and cleaned up a little bit.

PR #59, opened 2021-04-13
Improve CXL DVSEC decoding

bwidawsk:
These are some patches I've had sitting around for a while. They were used to help bring up the CXL driver in Linux 5.12.

### DIFF
--- a/ls-ecaps.c
+++ b/ls-ecaps.c
@@ -804,9 +804,46 @@ dvsec_cxl_port(uint8_t* data, int rev)
   printf("\t\tAlternateBus:\t%04x-%04x\n", m1, m2);
 }
 
+static const char *id[] = {
+  "empty",
+  "component registers",
+  "BAR virtualization",
+  "CXL device registers"};
+
+static inline void
+dvsec_decode_block(uint32_t lo, uint32_t hi, char which)
+{
+  u64 base_hi = hi, base_lo;
+  u8 bir, block_id;
+
+  bir = BITS(lo, 0, 3);
+  block_id = BITS(lo, 8, 8);
+  base_lo = BITS(lo, 16, 16);
+
+  if (!block_id)
+    return;
+
+  printf("\t\tBlock%c\tBIR: bar%d\tID: %s\n", which, bir, id[block_id]);
+  printf("\t\t\tRegisterOffset: %016" PCI_U64_FMT_X "\n", (base_hi << 32ULL) | base_lo << 16);
+}
+
+static void
+dvsec_cxl_register_locator(uint8_t* data, int len, int rev)
+{
+  int i, j;
+
+  if (rev != 0)
+    return;
+
+  for (i = 0xc, j = 1; i < len; i += 8, j++) {
+    dvsec_decode_block(*(u32 *)(data + i), *(u32 *)(data + i + 4), j + 0x31);
+  }
+}
+
 static void
 cap_dvsec_cxl(struct device *d, int id, int where)
 {
+  u16 len;
   u8 rev;
 
   printf(": CXL\n");
@@ -827,6 +864,13 @@ cap_dvsec_cxl(struct device *d, int id, int where)
         return;
 
       dvsec_cxl_port(d->config + where, rev);
+      break;
+    case 8:
+      len = BITS(get_conf_word(d, where + 0x6), 4, 12);
+      if (!config_fetch(d, where, len))
+        return;
+
+      dvsec_cxl_register_locator(d->config + where, len, rev);
       break;
     default:
       break;
--- a/ls-ecaps.c
+++ b/ls-ecaps.c
@@ -690,12 +690,15 @@ cap_rcec(struct device *d, int where)
 }
 
 static void
-cap_dvsec_cxl(struct device *d, int where)
+cap_dvsec_cxl(struct device *d, int id, int where)
 {
   u16 w;
 
   printf(": CXL\n");
   if (verbose < 2)
+    return;
+
+  if (id != 0)
     return;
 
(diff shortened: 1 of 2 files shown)


ls-ecaps.c
```diff
@@ -690,30 +690,202 @@ cap_rcec(struct device *d, int where)
...
 
-  printf(": CXL\n");
-  if (verbose < 2)
-    return;
+  w = (u16) size;
 
-  if (!config_fetch(d, where + PCI_CXL_CAP, 12))
+  size &= ~0x0fffffffULL;
+
+  printf("\t\tRange%d: %"PRIx64"-%"PRIx64"\n", n, base, base + size - 1);
```
pali, line 704: Just a small suggestion, what about adding left padding by zeros for range numbers? Like when printing base/limit numbers for BARs and windows.

pali, line 704: `PRIx64` would not work with windows msvcrt runtime. libpci has for this purpose its own format macro `PCI_U64_FMT_X`.

bwidawsk:
> Could you please rebase it on the current master?

Done

ls-ecaps.c
```diff
@@ -778,6 +778,28 @@ dvsec_cxl_device(uint8_t *data, int rev)
...
+  if (rev != 0)
+    return;
+
+  w = *(u16 *)(data + PCI_CXL_PORT_EXT_STATUS);
+  printf("\t\tCXLPortSta:\tPMComplete%c\n", FLAG(w, PCI_CXL_PORT_EXT_STATUS));
+
+  w = *(u16 *)(data + PCI_CXL_PORT_CTRL);
+  printf("\t\tCXLPortCtl:\tUnmaskSBR%c UnmaskLinkDisable%c AltMem%c AltBME%c\n",
+    FLAG(w, PCI_CXL_PORT_UNMASK_SBR), FLAG(w, PCI_CXL_PORT_UNMASK_LINK),
+    FLAG(w, PCI_CXL_PORT_ALT_MEMORY), FLAG(w, PCI_CXL_PORT_ALT_BME));
```
jic23, line 796: Why not report viral enable in bit 14?

ls-ecaps.c
```diff
@@ -778,6 +778,28 @@ dvsec_cxl_device(uint8_t *data, int rev)
...
+  u8 b1, b2;
+
+  if (rev != 0)
+    return;
+
+  w = *(u16 *)(data + PCI_CXL_PORT_EXT_STATUS);
+  printf("\t\tCXLPortSta:\tPMComplete%c\n", FLAG(w, PCI_CXL_PORT_EXT_STATUS));
+
+  w = *(u16 *)(data + PCI_CXL_PORT_CTRL);
+  printf("\t\tCXLPortCtl:\tUnmaskSBR%c UnmaskLinkDisable%c AltMem%c AltBME%c\n",
```
jic23, line 794: Use of tab after CXLPortCtl: here and not after AlternateBus: below leaves this block looking a bit oddly aligned.

ls-ecaps.c
```diff
@@ -778,6 +778,28 @@ dvsec_cxl_device(uint8_t *data, int rev)
...
+    FLAG(w, PCI_CXL_PORT_UNMASK_SBR), FLAG(w, PCI_CXL_PORT_UNMASK_LINK),
+    FLAG(w, PCI_CXL_PORT_ALT_MEMORY), FLAG(w, PCI_CXL_PORT_ALT_BME));
+
+  b1 = *(u8 *)(data + PCI_CXL_PORT_ALT_BUS_BASE);
+  b2 = *(u8 *)(data + PCI_CXL_PORT_ALT_BUS_LIMIT);
+  printf("\t\tAlternateBus: %02x-%02x\n", b1, b2);
+}
```
jic23, line 801: Nice to have the rest of this dvsec decoded, but that could be a follow up series...
* Alternative memory base + limit
* Alternative prefetchable base + limit.
presented like normal memory behind bridge entries...

ls-ecaps.c
bwidawsk: Tabs don't seem inconsistent to me. I will fix the missed tab though.

ls-ecaps.c
```diff
@@ -690,35 +690,49 @@ cap_rcec(struct device *d, int where)
...
     return;
 
-  w = get_conf_word(d, where + PCI_CXL_CAP);
+  w = *(u16 *)(data + PCI_CXL_CAP);
   printf("\t\tCXLCap:\tCache%c IO%c Mem%c Mem HW Init%c HDMCount %d Viral%c\n",
     FLAG(w, PCI_CXL_CAP_CACHE), FLAG(w, PCI_CXL_CAP_IO), FLAG(w, PCI_CXL_CAP_MEM),
     FLAG(w, PCI_CXL_CAP_MEM_HWINIT), PCI_CXL_CAP_HDM_CNT(w), FLAG(w, PCI_CXL_CAP_VIRAL));
 
-  w = get_conf_word(d, where + PCI_CXL_CTRL);
+  w = *(u16 *)(data + PCI_CXL_CTRL);
```
gollux: This does not work. Nobody guarantees that accesses through pointers to u16/u32 will have right endianity. Nor that they will not crash because of misalignment.

Please use `get_conf_xxx`, it's there for a reason.

bwidawsk: My bad. Will fix.

jic23:
Anything I can do to help this one along?  We are starting to see significant CXL devices being announced so would be great to have this basic support available (I have some stuff to add once Ben's initial series is upstream)

gollux:
Issues raised during review are still not fixed -- in particular, accessing cached config data by type-casting pointers. If you want, you can help clean up the code.

gollux:
Implemented in PR #110.

PR #64, opened 2021-05-18
Use POSIX shell variants

concatime:
For my first commit, it’s the same as https://github.com/NetworkConfiguration/dhcpcd/pull/38.
For the second, `command -v` is the POSIX way to check if a command exists.
Thank you.

### DIFF
--- a/lib/configure
+++ b/lib/configure
@@ -9,7 +9,7 @@ echo_n() {
 	printf '%s' "$*"
 }
 
-if [ -z "$VERSION" -o -z "$IDSDIR" ] ; then
+if [ -z "$VERSION" ] || [ -z "$IDSDIR" ] ; then
 	echo >&2 "Please run the configure script from the top-level Makefile"
 	exit 1
 fi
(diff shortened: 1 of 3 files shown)


JohnAZoidberg:
> For the second, command -v is the POSIX way to check if a command exists.

See https://github.com/pciutils/pciutils/pull/44

concatime:
`command` is optional only in [IEEE Std 1003.1, 2004 Edition](https://pubs.opengroup.org/onlinepubs/009604499/utilities/command.html). It’s optional in neither [IEEE Std 1003.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/utilities/command.html) nor the latest [IEEE Std 1003.1-2017 (Revision of IEEE Std 1003.1-2008)](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html).

gollux:
Is there any concrete system + shell combination where these changes are needed? Making source uglier just for sake of theoretical standard compliance is not wise.

pali:
Just my 2c, I have never seen any /bin/sh shell where `test` command did not support `-o` or `-a` switch. But in past when I played with Solaris I remember that `which` command was not present.

notorand-it:
My commit [here](https://github.com/pciutils/pciutils/pull/127) is not aiming at any compliancy but rather to make the script runnable under more shells, more notably under bash.

guillemj:
> Just my 2c, I have never seen any /bin/sh shell where `test` command did not support `-o` or `-a` switch. But in past when I played with Solaris I remember that `which` command was not present.

In Debian the `debianutils` maintainer (from where `which` is provided) intended to deprecate the `which` command, so switching to use `command -v` would be nice.

notorand-it:
> > Just my 2c, I have never seen any /bin/sh shell where `test` command did not support `-o` or `-a` switch. But in past when I played with Solaris I remember that `which` command was not present.
> 
> In Debian the `debianutils` maintainer (from where `which` is provided) intended to deprecate the `which` command, so switching to use `command -v` would be nice.

`command -v`  is not supported by all tools and most of the returns non-zero values, making it more difficult to discern from a 127 ("command not found", AKA `ENOENT`).

Instead, it's better for each tool to check whether a no-op option (like `--version`, `--help` etc.)

guillemj:
As mentioned on the other PR, `command -v` is an actual shell built-in, and not trying to randomly pass `-v` to external commands.

notorand-it:
> As mentioned on the other PR, `command -v` is an actual shell built-in, and not trying to randomly pass `-v` to external commands.

I see. OK. Just looking forward for this to be fixed: the script doesn't work with  `bash`.
What is needed for that PR (the one using `command -v`) to be merged?

concatime:
> Just my 2c, I have never seen any /bin/sh shell where `test` command did not support `-o` or `-a` switch. But in past when I played with Solaris I remember that `which` command was not present.

@pali https://git.suckless.org/sbase/ would prove you wrong.

PR #68, opened 2021-12-06
lib: check NULL before dereference

lixiaokeng:
udev_list_entry_get_name may fail and reutrn NULL.
Check it.

Signed-off-by:Lixiaokeng<lixiaokeng@huawei.com>


gollux:
Thanks, I will merge this.

gollux:
Solved a little bit differently to handle NULL returned by as the entry's value, too.

PR #69, opened 2021-12-07
lib: check malloc return value in pci_alloc

lixiaokeng:
The malloc may failed and return NULL. Check it
before using.

Signed-off-by: Lixiaokeng<lixiaokeng@huawei.com>


gollux:
Thanks, I will merge this.

gollux:
I chose a different solution, but fixed.

PR #70, opened 2021-12-07
ls-map: fix memory leak in do_map_bus

lixiaokeng:
The struct device "d" is freed without freeing its
pointer fileds "config" and "present".

Signed-off-by:Lixiaokeng <lixiaokeng@huawei.com>


gollux:

PR #71, opened 2021-12-07
lspci: fix memory leak in scan_device

lixiaokeng:
If return NULL, we should free d and d's member.

Signed-off-by: Lixiaokeng <lixiaokeng@huawei.com>


gollux:

PR #72, opened 2021-12-07
ls-kernel: fix memory leak in show_kernel_init

lixiaokeng:
The pointer e will not be free if continue. Fix it.

Signed-off-by:Lixiaokeng <lixiaokeng@huawei.com>


gollux:

PR #74, opened 2021-12-25
libpci: Improve i386-io-windows.h port

pali:
This pull request improves i386-io-windows.h code:
* Fix support for pre-NT systems (no special code is needed, access to PCI I/O ports is always allowed).
* Fix compilation with older CRTDLL and new modern UCRT runtime libraries.
* Replace usage of license incompatible WinIO.dll library and requirement of having installed insecure 3rd-party kernel module WinIo.sys by NtSetInformationProcess(ProcessUserModeIOPL) syscall supported directly by NT kernels (equivalent of iopl(3) on Linux). Issuing NtSetInformationProcess(ProcessUserModeIOPL) syscall requires to have special NT privileges and this change also implements temporary privilege grants for the calling thread according to token access guidelines.

Differently compiled lspci.exe binaries were tested in VM on NT 3.51, NT 4.0, 98 SE, 2000 and XP under Administrator account.

### DIFF
--- a/lib/i386-io-windows.h
+++ b/lib/i386-io-windows.h
@@ -3,12 +3,14 @@
  *
  *	Copyright (c) 2004 Alexander Stock <stock.alexander@gmx.de>
  *	Copyright (c) 2006 Martin Mares <mj@ucw.cz>
+ *	Copyright (c) 2021 Pali Rohár <pali@kernel.org>
  *
  *	Can be freely distributed and used under the terms of the GNU GPL.
  */
 
 #include <io.h>
 #include <windows.h>
+#include <aclapi.h>
 
 #ifdef _MSC_VER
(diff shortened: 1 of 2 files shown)


lib/i386-io-windows.h
```diff
@@ -97,13 +99,1175 @@ unsigned long _inpd(unsigned short port);
...
+  unsigned
+#ifdef __x86_64__
+  long long
+#endif
+  int eflags;
+  asm volatile ("pushf\n\tpop %0\n" : "=r" (eflags));
+  return eflags;
+}
+#else
+#error Unsupported compiler
```
gollux, line 145: The argument of `#error` should be a string.

pali: fixed

PR #75, opened 2021-12-26
Fix compilation of source files with MinGW

pali:
Improve compat/getopt.c implementation, fix usage of vsnprintf() and strcasecmp() functions with MSVCRT runtime and fix compile errors.

### DIFF
--- a/ls-ecaps.c
+++ b/ls-ecaps.c
@@ -653,8 +653,9 @@ cap_rcec(struct device *d, int where)
       int prevmatched=0;
       int adjcount=0;
       int prevdev=0;
+      int dev;
       printf("RCiEP at Device(s):");
-      for (int dev=0; dev < 32; dev++)
+      for (dev=0; dev < 32; dev++)
         {
 	  if (BITS(bmap, dev, 1))
 	    {
--- a/ls-tree.c
+++ b/ls-tree.c
@@ -170,6 +170,8 @@ tree_printf(char *line, char *p, char *fmt, ...)
   int res = vsnprintf(p, end - p, fmt, args);
   if (res < 0)
     {
+      /* msvcrt _vsnprintf() does not set nul-term byte */
+      *end = '\0';
       /* Ancient C libraries return -1 on overflow */
       p += strlen(p);
     }
--- a/compat/getopt.c
+++ b/compat/getopt.c
@@ -36,7 +36,6 @@
 #endif
 
 #include <stdio.h>
-#include <string.h>
 
 /* Comment out all this code if we are using the GNU C Library, and are not
    actually compiling the library itself.  This code is part of the GNU C
@@ -163,6 +162,8 @@ static enum {
 #include <string.h>
 #define	my_index	strchr
 #define	my_strlen	strlen
+#define	my_strcmp	strcmp
+#define	my_strncmp	strncmp
 #else
 
 /* Avoid depending on library functions or files
@@ -170,11 +171,11 @@ static enum {
 
 #if __STDC__ || defined(PROTO)
 extern char *getenv(const char *name);
-extern int strcmp(const char *s1, const char *s2);
-extern int strncmp(const char *s1, const char *s2, int n);
 
 static int my_strlen(const char *s);
 static char *my_index(const char *str, int chr);
+static int my_strncmp(const char *s1, const char *s2, int n);
+static int my_strcmp(const char *s1, const char *s2);
 #else
 extern char *getenv();
 #endif
@@ -195,6 +196,23 @@ static char *my_index(const char *str, int chr)
 		str++;
 	}
 	return 0;
+}
+
+static int my_strncmp(const char *s1, const char *s2, int n)
+{
+	while (n && *s1 && (*s1 == *s2)) {
+		++s1;
+		++s2;
+		--n;
+	}
+	if (n == 0)
+		return 0;
+	return *(const unsigned char *)s1 - *(const unsigned char *)s2;
+}
+
+static int my_strcmp(const char *s1, const char *s2)
+{
+	return my_strncmp(s1, s2, -1);
 }
 
 #endif				/* GNU C library.  */
@@ -385,7 +403,7 @@ int _getopt_internal(int argc, char *const *argv, const char *optstring,
 		   then exchange with previous non-options as if it were an option,
 		   then skip everything else like a non-option.  */
 
-		if (optind != argc && !strcmp(argv[optind], "--")) {
+		if (optind != argc && !my_strcmp(argv[optind], "--")) {
 			optind++;
 
 			if (first_nonopt != last_nonopt && last_nonopt != optind)
@@ -445,7 +463,7 @@ int _getopt_internal(int argc, char *const *argv, const char *optstring,
 
 		/* Test all options for either exact match or abbreviated matches.  */
 		for (p = longopts, option_index = 0; p->name; p++, option_index++)
-			if (!strncmp(p->name, nextchar, s - nextchar)) {
+			if (!my_strncmp(p->name, nextchar, s - nextchar)) {
 				if (s - nextchar == my_strlen(p->name)) {
 					/* Exact match found.  */
 					pfound = p;
--- a/compat/getopt.h
+++ b/compat/getopt.h
@@ -95,12 +95,7 @@ struct option
 #define optional_argument	2
 
 #if __STDC__ || defined(PROTO)
-#if defined(__GNU_LIBRARY__)
-/* Many other libraries have conflicting prototypes for getopt, with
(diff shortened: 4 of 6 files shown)


ls-ecaps.c
```diff
@@ -653,8 +653,9 @@ cap_rcec(struct device *d, int where)
       int prevmatched=0;
       int adjcount=0;
       int prevdev=0;
+      int dev;
       printf("RCiEP at Device(s):");
-      for (int dev=0; dev < 32; dev++)
+      for (dev=0; dev < 32; dev++)
```
gollux, line 658: The error message suggests that the compiler supports C99 mode, so turn it on please.

Years ago, I preferred to avoid C99 in the pciutils, but that times are long gone. It's 22 years since C99 was released, so it's time to say farewall to compilers which still do not support it.

ls-tree.c
```diff
@@ -170,6 +170,8 @@ tree_printf(char *line, char *p, char *fmt, ...)
   int res = vsnprintf(p, end - p, fmt, args);
   if (res < 0)
     {
+      /* msvcrt _vsnprintf() does not set nul-term byte */
+      *end = '\0';
```
gollux, line 174: This fixes one call to (v)snprintf, but leaves others unfixed. On platforms with broken snprintf, please provide a wrapper instead.

gollux: When doing that, please move the "returns -1 instead of proper length" case to the wrapper, too.

ls-ecaps.c
pali: Ah, stupid that it is disabled by default... With `CFLAGS="-std=gnu99"` this one patch is not needed.

ls-tree.c
pali: After inspection I realized that this is the only (v)snprintf on windows platform. All other files with (v)snprintf calls are not used on windows platform.

gollux: What about those in `lib/names.c`, for example?

Still, I do not want random platform-specific hacks all over the code.

pali: I see, I will look at it...

pali: I'm looking into `lib/names.c` but I do not understand what should `format_name()` do when overflow occurs. It contains this code:

```c
    res = snprintf(buf, size, "%s [%s]", name, num);
  if (res >= size && size >= 4)
    buf[size-2] = buf[size-3] = buf[size-4] = '.';
  else if (res < 0 || res >= size)
    return "<pci_lookup_name: buffer too small>";
  return buf;
```

It fills ellipsis into target buffer and then returns different buffer. So it is required to touch buffer if overflow occurs or not?

gollux: Does it really return a different buffer? Maybe you missed the `else`?

pali: Ou right there is `else` branch... (it is early morning, not so good time for thinking :))

pali: Ok, its not early morning anymore... I'm looking at it again and it remind me that I was traversing this code months ago. And really no other usage of (v)snprintf except this one in `tree_printf` needs special handling.

The reason is because all other non-platform-specific calls of (v)snprintf completely drops buffer when negative value is returned. The only exception is  `tree_printf` in `ls-tree.c`.

C11 standard about `vsnprintf` in 7.21.6.12 says:
> The `vsnprintf` function returns the number of characters that would have been written had `n` been sufficiently large, not counting the terminating null character, or a negative value if an encoding error occurred. Thus, the null-terminated output has been completely written if and only if the returned value is nonnegative and less than `n`.

In my opinion caller of `vsnprintf` should not expect anything about what is stored in buffer when `vsnprintf` returns negative value.

Should not also in this case `tree_printf` fill `p` buffer by some `<buffer too small>` filler when `vsnprintf` returns negative value? It would make code more compliant with standard and also elegantly fixes this problem that _msvcrt vsnprintf does not fill nul-term byte when it returns negative value_.

gollux: `vsnprintf` is defined as a variant of `snprintf`, whose definition in section 7.21.6.5 (since C99) is more explicit with respect to what happens when the output is too large:

> The `snprintf` function is equivalent to `fprintf`, except that the output is written into
an array (specified by argument `s`) rather than to a stream. If `n` is zero, nothing is written,
and s may be a null pointer. Otherwise, output characters beyond the `n-1`st are
discarded rather than being written to the array, and a null character is written at the end
of the characters actually written into the array. [...]

So *msvcrt* is the guilty party here.

pali: > So _msvcrt_ is the guilty party here.

Of course. But negative value from *printf() family is allowed. And my question is how to handle it?

pali: So, what about this approach?

```diff
diff --git a/ls-tree.c b/ls-tree.c
index ba7873b77649..0ad8567b7fc5 100644
--- a/ls-tree.c
+++ b/ls-tree.c
@@ -167,17 +167,28 @@ tree_printf(char *line, char *p, char *fmt, ...)
 {
   va_list args;
   char *end = line + LINE_BUF_SIZE - 2;
+  int size = end - p;
+  int res;
 
-  if (p >= end)
+  if (size <= 0)
     return p;
 
   va_start(args, fmt);
-  int res = vsnprintf(p, end - p, fmt, args);
-  if (res < 0)
+  res = vsnprintf(p, size, fmt, args);
+  if (res >= size && size >= 4)
+    p[size-2] = p[size-3] = p[size-4] = '.';
+  else if (res < 0 || res >= size)
     {
-      /* Ancient C libraries return -1 on overflow */
-      p += strlen(p);
+      /*
+       * vsnprintf() returns a negative value if error occurred.
+       * Ancient C libraries return -1 on overflow.
+       */
+      memset(p, '.', size-1);
+      p[size-1] = '\0';
     }
+
+  if (res >= size)
+    p += size-1;
   else
     p += res;
 
```

Note that now I spotted another issue in `tree_printf()` function that it returns pointer `p` increased by `res` also when C99-style overflow occurred. Return value of `tree_printf()` is then passed to `print_it()` which calls `*p++ = '\n'` and writes out of `line[]` array.

gollux: Well, a negative result is technically allowed, but only for "encoding error". And even though it is hard to tell what it means, it most certainly does not mean "buffer too small". But given how hard it is to eradicate all evil in the world, we have to work around it...

However, filling the whole buffer with dots makes a little sense...

I committed a different solution, can you have a look at it, please?

pali: Looks good!

gollux:
Great! I cherry-picked the rest of this PR to my master branch. Could you please check that it compiles on MinGW?

pali:
There is still problem with for-loop because i586-mingw32msvc-gcc compiler has disabled C99 mode by default. But with hacking Makefile it finally compiles.

Second issue is that compiler throws lot of printf format warnings like:
* lspci.c:453: warning: format ‘%04x’ expects type ‘unsigned int’, but argument 2 has type ‘pciaddr_t’
* lspci.c:518: warning: format ‘%08x’ expects type ‘unsigned int’, but argument 2 has type ‘long unsigned int’
* lspci.c:571: warning: format ‘%x’ expects type ‘unsigned int’, but argument 2 has type ‘u32’
* lspci.c:670: warning: format ‘%08x’ expects type ‘unsigned int’, but argument 3 has type ‘u32’
* ls-caps.c:101: warning: format ‘%d’ expects type ‘int’, but argument 2 has type ‘long unsigned int’
* ...

pali:
I have looked that those format warnings and they are reall issues (not just warnings), which makes compiled lspci binary not fully working. MSVC in those cases throw fatal errors correctly.

Fixes for these runtime issues are already in git branch linked to pull request https://github.com/pciutils/pciutils/pull/78 and because this pull request is already merged I'm closing it.

ls-tree.c
pali: > Well, a negative result is technically allowed, but only for "encoding error". And even though it is hard to tell what it means, it most certainly does not mean "buffer too small".

I guess that "encoding error" refers to printing wide strings or similar cases...

Anyway I sent patches to mingw-w64 which provide sane (v)snprintf() wrapper around msvcrt _(v)snprintf(). Trailing nul-term byte would be correctly filled also on overflow and required length would be calculated on overflow as defined in C99. But there are still at least 2 cases when negative value value by (v)snprintf() would be returned: 1. msvcrt signals that format string is invalid or 2. malloc() used by that wrapper returns NULL.

So negative value returned by (v)snprintf() always needs to be handled and no assumption should be done about what is stored in buffer in this case.

Btw, I figured out that this problem has already assigned CVE number: [CVE-2018-1000101](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1000101)

PR #76, opened 2021-12-26
Fix configure and Makefile for MinGW

pali:
Current Makefile does not work with MinGW cross compilers. Fix it together with configure script for cross compilation.


gollux (changes requested):
Generally, it looks OK. The `$(EXEEXT)` part is ugly as hell, but probably unavoidable.

The command used to compile pciutils on Windows should better be written in `README.Windows` instead of hiding it inside commit messages :)

pali:
Done! I rebased code on top of the master branch and added new commit with compile instruction.

gollux:
Could you please rebase it again against the current master?

pali:
Done again!

PR #77, opened 2021-12-26
libpci: For PCI_OS_WINDOWS allows to load pci.ids from executable directory

pali:
With compile option PCI_PATH_IDS_DIR="" loads pci.ids on Windows from the executable directory.

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -13,6 +13,10 @@
 
 #include "internal.h"
 
+#ifdef PCI_OS_WINDOWS
+#include <windows.h>
+#endif
+
 static struct pci_methods *pci_methods[PCI_ACCESS_MAX] = {
   NULL,
 #ifdef PCI_HAVE_PM_LINUX_SYSFS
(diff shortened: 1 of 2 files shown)


lib/init.c
```diff
@@ -187,7 +191,25 @@ pci_alloc(void)
   int i;
 
   memset(a, 0, sizeof(*a));
+#ifdef PCI_OS_WINDOWS
+  if ((PCI_PATH_IDS_DIR)[0])
+    pci_set_name_list_path(a, PCI_PATH_IDS_DIR "\\" PCI_IDS, 0);
+  else
+    {
+      char *path = malloc(MAX_PATH+1);
+      DWORD len = GetModuleFileNameA(NULL, path, MAX_PATH+1);
```
gollux, line 200: Crashes when `malloc` returns `NULL`. Use `pci_malloc` instead?

pali: `pci_malloc` cannot be used at this place as `a->error` (used by `pci_malloc`) is not initialized yet.

pali: fixed

gollux: I don't want to have many copies of an out of memory message. But is there any reason for `malloc`ing the buffer instead of having it as a local variable?

pali: `pci_set_name_list_path()` takes pointer of buffer for later usage, so `path` needs to be either global/static variable or `malloc`ed buffer. `pci_set_name_list_path()` API is designed for `malloc`ed buffers (third argument specify if libpci will call `free()`), so I chose to use `malloc()`. Do you prefer something else (static or global variable)?

gollux: Ah yes...

I changed `pci_malloc()`, so that it can be called even if `pci_access` is not fully initialized yet. You can now use it here.

pali: done!

gollux:
I cleaned it up a bit, could you please check that I broke nothing?

pali:
Checked and still works fine.

PR #82, opened 2021-12-31
Add native windows implementation via cfgmgr32.dll library

pali:
Access via cfgmgr32.dll library allows to list PCI devices and retrieve their basic properties and system resource configuration. Access is available to all users and should not require special privileges, access tokens, rights or permissions.

It works for both 32-bit and 64-bit systems. But there is a caveat, 32-bit application on 64-bit system cannot retrieve system resources.

Disadvantage of cfgmgr32.dll library that it does not provide access to PCI config space and so everything needs to be implemented via libpci properties.

Git branch linked to this pull request contains commits which extends libpci API/ABI for exporting useful PCI properties. I have also changes for lspci to use them and also implementations for sysfs and procfs providers, but these changes depends on other pending patches on linux-pci mailing list and github does not allow me to easily create a pull request which depends on these changes. So I pushed into branch linked to this pull request only changes to libpci API and then windows cfgmgr32 implementation. (I can provide remaining lspci non-windows patches here as attachments, if needed)

File win32-cfgmgr32.c compiles fine with MinGW32, MinGW-w64-32bit, MinGW-w64-64bit and MSVC compilers.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -87,6 +87,7 @@ i386-ports.o: i386-ports.c $(INCL) i386-io-hurd.h i386-io-linux.h i386-io-sunos.
 proc.o: proc.c $(INCL) pread.h
 sysfs.o: sysfs.c $(INCL) pread.h
 generic.o: generic.c $(INCL)
+emulated.o: emulated.c $(INCL)
 syscalls.o: syscalls.c $(INCL)
 obsd-device.o: obsd-device.c $(INCL)
 fbsd-device.o: fbsd-device.c $(INCL)
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -54,6 +54,11 @@ ifdef PCI_HAVE_PM_HURD_CONF
 OBJS += hurd
 endif
 
+ifdef PCI_HAVE_PM_WIN32_CFGMGR32
+OBJS += emulated
+OBJS += win32-cfgmgr32
+endif
+
 all: $(PCILIB) $(PCILIBPC)
 
 ifeq ($(SHARED),no)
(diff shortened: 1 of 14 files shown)


gollux:
This looks interesting, but I do not see where it is heading. So far, all backends provide access to configuration space and users of libpci rely on that. Do you plan to add a back-end, which provides only some device properties, but no config space access? Which clients are supposed to use that?

pali:
I have changes for lspci to use these PCI_FILL_* properties (when available) and allow lspci to work without access to config space... so lspci can be a user for it.

And new backend which provides these PCI_FILL_* properties without config access is already in this pull request in the last commit.

gollux:
After several hours of pondering over this, I still do not like this direction. You are essentially changing libpci API in a backwards-incompatible way: old clients of the library will not work with your back-end.

I suggest another approach: provide virtual config space, which will contain all information you can gather.

pali:
> I suggest another approach: provide virtual config space, which will contain all information you can gather.

Ok. Is simple 64-byte long virtual read-only config space enough?

gollux:
Yes.

pali:
Done!

lib/Makefile
```diff
@@ -101,3 +105,11 @@ names-hwdb.o: names-hwdb.c $(INCL) names.h
 filter.o: filter.c $(INCL)
 nbsd-libpci.o: nbsd-libpci.c $(INCL)
 hurd.o: hurd.c $(INCL)
+win32-cfgmgr32.o: win32-cfgmgr32.c $(INCL)
+
+# MinGW32 toolchain has some required Win32 header files in /ddk subdirectory.
+# But these header files includes another header files from /ddk subdirectory
+# and expects that build system already set /ddk subdirectory into includes.
+# So include /ddk subdirectory of each system predefined include path via -I.
+DDKCFLAGS:=$(shell echo | $(CC) $(CFLAGS) -E -Wp,-v -o /dev/null - 2>&1 | sed -n 's/^ \(.*\)/-I\1\/ddk/p')
```
gollux, line 114: Please hide the shell call behind an ifdef.

gollux:
When we are emulating the config space, I do not see any need for the extensions to `pci_fill_info`. They just complicate things unnecessarily.

pali:
Pull request is updated, fixed and rebased on master.

pali:
Updated, emulated virtual config space functions are now in separate emulated.c file and checks in lspci.c were slightly updated to not use config_cached.

PR #86, opened 2022-01-06
lspci: Show hex dump parsed with -X option.

AlbertDM:
New option -X added: 

    lspci -X

Based on -x option, but parses the hexadecimal dump following the PCIe Config Types standards.
This is intended for PCI development or board bring-up purposes.

### DIFF
--- a/lspci.c
+++ b/lspci.c
@@ -17,6 +17,7 @@
 
 int verbose;				/* Show detailed information */
 static int opt_hex;			/* Show contents of config space as hexadecimal numbers */
+static int opt_hex_parsed;	/* Show parsed contents of config space as hexadecimal numbers */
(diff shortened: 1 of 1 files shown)


gollux:
I would appreciate an explanation of why this output format is interesting (and preferred to infinitely many other possibilities).

AlbertDM:
Hello Martin,

For my current work I needed an option to help me represent and understand how the configuration space is transmitted from the EP to the RC based on the PCIe protocol, since I am working on the bring-up of new PCIe devices that not always work fine. With this new option I expected to be quicker to identify any value miss-match on a specific field.

I humbly apologize if this was already implemented, since I was not aware. Could you please show me similar solutions? This would help me on my current tasks.

Thanks!

PR #94, opened 2022-04-05
Fix lspci: PMCSR PCI to PCI Bridge Support Extensions

pepyaka:
Bridge PM and B3 were from "Power Management Control/Status Register", but should be from "PMCSR_BSE Bridge Support Extensions register"


gollux:
Thanks, applied.

PR #100, opened 2022-05-30
lib: Fix PCI_L1PM_SUBSTAT_CAP_L1PM_SUPP

pepyaka:
Hi!
The mask should be 0b10000, not 0b10110
Screenshot from 2022-05-30 18-27-20


gollux:
Thanks, applied!

PR #101, opened 2022-06-02
lspci: Fix VC Resource Capability Register / Maximum Time Slots

pepyaka:
Hi!
`VC Resource Capability Register / Maximum Time Slots` should be 7 bits long not 6
Screenshot from 2022-06-02 14-24-52


gollux:
Thanks, applied!

PR #104, opened 2022-06-16
setpci: Check if standard register on device exist

pali:
Some standard registers are available only on device with header type 0,
some only on header type 1, some other only on header type 2 and some on
header type 0 and 1. Add definitions which registers are available on which
header type and add check to access only available registers.


setpci.c
```diff
@@ -271,60 +281,60 @@ static const struct reg_name pci_reg_names[] = {
...
-  {       0, 0x30, 2, "CB_IO_LIMIT_0" },
-  {       0, 0x32, 2, "CB_IO_LIMIT_0_HI" },
-  {       0, 0x34, 2, "CB_IO_BASE_1" },
-  {       0, 0x36, 2, "CB_IO_BASE_1_HI" },
-  {       0, 0x38, 2, "CB_IO_LIMIT_1" },
-  {       0, 0x3a, 2, "CB_IO_LIMIT_1_HI" },
-  {       0, 0x40, 2, "CB_SUBSYSTEM_VENDOR_ID" },
-  {       0, 0x42, 2, "CB_SUBSYSTEM_ID" },
-  {       0, 0x44, 4, "CB_LEGACY_MODE_BASE" },
+  {       0, 0x10, 4, "BASE_ADDRESS_0", 0x3 },
```
gollux, line 284: This breaks alignment of table columns. Please put the type mask before the name.

gollux:
Nice, I will merge it when you fix the table alignment.

setpci.c
pali: Done, updated.

PR #105, opened 2022-06-19
lib/configure: drop usage of which

thesamesam:
'which' is not required by POSIX and is an external command which may not be
available, and 'command -v' does the job just fine.

Debian and Gentoo at least are both making efforts to drop which from
their base system package list.

Signed-off-by: Sam James <sam@gentoo.org>


gollux:
I consider dropping `which` by distributions quite silly, because `which` is often used and much easier to type than `command -v`.

Still, there is no harm in accepting the change...

thesamesam:
I understand, although it's scripted use, so it's not so bad there. Thanks!

PR #106, opened 2022-06-20
libpci: Add Intel Type 1 implementation for memory mapped systems

pali:
Lot of non-x86 platforms also support Intel Type 1 mechanism. x86 IO ports
CF8 and CFC are on these platforms mapped into standard memory space.
Address mapping itself is platform or board specific and there is no
default value.

Lot of ARM boards with multiple PCIe controllers are multi-domain and each
PCI domain has its own CF8/CFC (address/data) registers mapped into memory
space.

Add new mmio-conf1 backend which access CF8/CFC ports via MMIO and define
new config option mmio-conf1.addrs which specify list of address/data
register pairs in memory space for each PCI domain. Format of this option
is: 0xaddr1/0xdata1,0xaddr2/0xdata2,...

### DIFF
--- a/lib/configure
+++ b/lib/configure
@@ -68,12 +68,14 @@ LSPCIDIR=SBINDIR
 
 case $sys in
 	linux*)
-		echo_n " sysfs proc"
+		echo_n " sysfs proc mem-ports"
 		echo >>$c '#define PCI_HAVE_PM_LINUX_SYSFS'
 		echo >>$c '#define PCI_HAVE_PM_LINUX_PROC'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
 		echo >>$c '#define PCI_HAVE_LINUX_BYTEORDER_H'
 		echo >>$c '#define PCI_PATH_PROC_BUS_PCI "/proc/bus/pci"'
 		echo >>$c '#define PCI_PATH_SYS_BUS_PCI "/sys/bus/pci"'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		case $cpu in
 				i?86|x86_64)	echo_n " i386-ports"
 						echo >>$c '#define PCI_HAVE_PM_INTEL_CONF'
@@ -97,7 +99,9 @@ case $sys in
 	freebsd*|kfreebsd*)
 		echo_n " fbsd-device"
 		echo >>$c '#define PCI_HAVE_PM_FBSD_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
 		echo >>$c '#define PCI_PATH_FBSD_DEVICE "/dev/pci"'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		if [ "$sys" != "kfreebsd" ] ; then
 			LIBRESOLV=
 		fi
@@ -105,13 +109,17 @@ case $sys in
         openbsd)
 	        echo_n " obsd-device"
 		echo >>$c '#define PCI_HAVE_PM_OBSD_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
 		echo >>$c '#define PCI_PATH_OBSD_DEVICE "/dev/pci"'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		LIBRESOLV=
 		;;
 
         darwin*)
 	        echo_n " darwin"
 		echo >>$c '#define PCI_HAVE_PM_DARWIN_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		echo >>$m 'WITH_LIBS+=-lresolv -framework CoreFoundation -framework IOKit'
 		echo >>$c '#define PCI_HAVE_64BIT_ADDRESS'
 		LIBRESOLV=
@@ -121,6 +129,8 @@ case $sys in
 	aix)
 		echo_n " aix-device"
 		echo >>$c '#define PCI_HAVE_PM_AIX_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		echo >>$m 'CFLAGS=-g'
 		echo >>$m 'INSTALL=installbsd'
 		echo >>$m 'DIRINSTALL=mkdir -p'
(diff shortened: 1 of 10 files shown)


gollux:
I like the idea, but I have a couple of comments:

- The access method should not have the same name as pre-existing `intel-conf1`. Let's call it for example `mm-conf1` and use the same name for the source file and option names (`mm-conf1.addrs` and `mm-conf1.devmem`).
- Does it have to be mutually exclusive with `intel-conf`?
- Please add documentation to `pcilib.man`. Make it clear that this access method must be configured before it works.
- Is it necessary to `munmap` the memory region after each access?

pali:
> * The access method should not have the same name as pre-existing `intel-conf1`.

I called it `intel-conf1` because it is same access method, just used on different platforms / architectures. Same as existing different implementations of `intel-conf1` for different operating systems. That is the idea behind.

> * Does it have to be mutually exclusive with `intel-conf`?

Yes and no? On x86 it is possible to compile both methods but only the IO-based can work. On non-x86 systems it is possible to compile only MEM-based. I think it does not make sense to compile MEM-based for x86 (as it would not work) and therefore those two methods would be mutually exclusive.

Or is there some x86 PCI controller which has CF8/CFC registers remapped from IO to memory space, and which can be accessed by `mov` instruction?

> * Please add documentation to `pcilib.man`. Make it clear that this access method must be configured before it works.

Ok!

> * Is it necessary to `munmap` the memory region after each access?

On multi-PCIe segment systems (which is every ARM SoC with more than one PCIe serdes), each segment has its own CF8/CFC registers. So to access selected domain it is needed to map correct address. Calling `munmap` is there just as simplification, no need to store mapped memory pointer for each segment in some internal data structure.

gollux:
> I called it intel-conf1 because it is same access method, just used on different platforms / architectures. Same as existing different implementations of intel-conf1 for different operating systems. That is the idea behind.

I see... but I still think that it is substantially different, especially as it has to be configured in order to work.

> Yes and no? On x86 it is possible to compile both methods but only the IO-based can work. On non-x86 systems it is possible to compile only MEM-based. I think it does not make sense to compile MEM-based for x86 (as it would not work) and therefore those two methods would be mutually exclusive.

I do not think that it matters that no current machine supports both access methods. Technically, the two access methods do not collide with each other, so I would like to make it possible to enable both.

> On multi-PCIe segment systems (which is every ARM SoC with more than one PCIe serdes), each segment has its own CF8/CFC registers. So to access selected domain it is needed to map correct address. Calling munmap is there just as simplification, no need to store mapped memory pointer for each segment in some internal data structure.

I shiver when I imagine the storm of `mmap`s and `munmap`s when running plain `lspci` :)

Please cache the `mmap`ed window for the most-recently accessed device.

pali:
Ok! I will do it.

pali:
mmap caching implemented, documentation added and backend renamed, so now it is not in conflict with intel one.

Tested on big endian PowerPC P2020 CPU which has 3 PCIe controllers (domains) where each controller has own CF8/CFC I/O ports mapped to memory space and to each controller is connected PCIe device:
```
# lspci -A linux-sysfs -nn
0000:04:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0000:05:00.0 Network controller [0280]: Qualcomm Atheros QCA986x/988x 802.11ac Wireless Network Adapter [168c:003c]
0001:02:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0001:03:00.0 Network controller [0280]: Qualcomm Atheros AR93xx Wireless Network Adapter [168c:0030] (rev 01)
0002:00:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0002:01:00.0 USB controller [0c03]: Texas Instruments TUSB73x0 SuperSpeed USB 3.0 xHCI Host Controller [104c:8241] (rev 02)
```
```
# strace lspci -A mmio-conf1 -O mmio-conf1.addrs=0xffe0a000/0xffe0a004,0xffe09000/0xffe09004,0xffe08000/0xffe08004 -nn 3>&2 2>&1 1>&3 | grep -i map
mmap2(NULL, 4096, PROT_READ|PROT_WRITE, MAP_SHARED, 4, 0xffe0a000) = 0xb7a9d000
mmap2(NULL, 4096, PROT_READ|PROT_WRITE, MAP_SHARED, 4, 0xffe09000) = 0xb7a9c000
munmap(0xb7a9d000, 4096)                = 0
mmap2(NULL, 4096, PROT_READ|PROT_WRITE, MAP_SHARED, 4, 0xffe08000) = 0xb7a9d000
munmap(0xb7a9c000, 4096)                = 0
mmap2(NULL, 4096, PROT_READ|PROT_WRITE, MAP_SHARED, 4, 0xffe09000) = 0xb7a9c000
munmap(0xb7a9d000, 4096)                = 0
mmap2(NULL, 4096, PROT_READ|PROT_WRITE, MAP_SHARED, 4, 0xffe0a000) = 0xb7a9d000
munmap(0xb7a9c000, 4096)                = 0
0000:00:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0000:05:00.0 Network controller [0280]: Qualcomm Atheros QCA986x/988x 802.11ac Wireless Network Adapter [168c:003c]
0001:00:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0001:03:00.0 Network controller [0280]: Qualcomm Atheros AR93xx Wireless Network Adapter [168c:0030] (rev 01)
0002:00:00.0 PCI bridge [0604]: Freescale Semiconductor Inc P2020E [1957:0070] (rev 21)
0002:01:00.0 USB controller [0c03]: Texas Instruments TUSB73x0 SuperSpeed USB 3.0 xHCI Host Controller [104c:8241] (rev 02)
munmap(0xb7a9d000, 4096)                = 0
```

lib/configure
```diff
@@ -97,21 +99,27 @@ case $sys in
...
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
 		echo >>$c '#define PCI_PATH_OBSD_DEVICE "/dev/pci"'
+		echo >>$c '#define PCI_PATH_DEVMEM_DEVICE "/dev/mem"'
 		LIBRESOLV=
 		;;
 
         darwin*)
 	        echo_n " darwin"
 		echo >>$c '#define PCI_HAVE_PM_DARWIN_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
```
gollux, line 121: Does it really make sense to use this on Darwin? Is there an Apple machine where it can be used?

lib/configure
```diff
@@ -121,14 +129,18 @@ case $sys in
 	aix)
 		echo_n " aix-device"
 		echo >>$c '#define PCI_HAVE_PM_AIX_DEVICE'
+		echo >>$c '#define PCI_HAVE_PM_MMIO_CONF'
```
gollux, line 132: Similarly for AIX. I suspect that no new AIX machines will ever appear...

gollux (changes requested):
Otherwise, I agree.

pali:
Memory mapped CF8/CFC access is heavily used on ARM but I found also some PowerPC processor on which I tested it. I was surprised that some PowerPC has it. And I have not seen any other platform which have this pci config space access.

In past Apple used PowerPC, x86 and now Aarch64. I was told that their new ARM processors uses standard PCIe ECAM config space access method. I bet that their x86 does not have memory mapped CF8/CFC. And about their old PowerPC I'm not sure.

About AIX system, currently it supports POWER9 and in past it was running also on x86 and older PowerPC architecture. I do not know neither.

It is like for x86, I do not see a big benefit to have it enabled for x86. It can be compiled without issue but there is probably no such HW which can use it.

PR #109, opened 2022-08-19
hurd: fix filling information

sthibaul:
The hurdish methods only implement the region and base information, the
rest should be taken from the generic method.


gollux:
Thanks!

PR #110, opened 2022-09-16
lspci: Improve CXL DVSEC decoding

jphaws:
I wanted to help #59 along since it's been open for awhile and I'm interested in following up with additional CXL DVSEC decoding. I addressed the accessing cached config data by type-casting pointers issues by replacing them with their respective `get_conf_xxx` and rebased against the current master. Other than that, everything remains as originally written by @bwidawsk.

### DIFF
--- a/ls-ecaps.c
+++ b/ls-ecaps.c
@@ -690,8 +690,30 @@ cap_rcec(struct device *d, int where)
 }
 
 static void
+cxl_range(u64 base, u64 size, int n)
+{
+  u32 interleave[] = { 0, 256, 4096, 512, 1024, 2048, 8192, 16384 };
+  const char *type[] = { "Volatile", "Non-volatile", "CDAT" };
+  const char *class[] = { "DRAM", "Storage", "CDAT" };
+  u16 w;
+
+  w = (u16) size;
+
+  size &= ~0x0fffffffULL;
+
+  printf("\t\tRange%d: %016"PCI_U64_FMT_X"-%016"PCI_U64_FMT_X"\n", n, base, base + size - 1);
+  printf("\t\t\tValid%c Active%c Type=%s Class=%s interleave=%d timeout=%ds\n",
+    FLAG(w, PCI_CXL_RANGE_VALID), FLAG(w, PCI_CXL_RANGE_ACTIVE),
+    type[PCI_CXL_RANGE_TYPE(w)], class[PCI_CXL_RANGE_CLASS(w)],
+    interleave[PCI_CXL_RANGE_INTERLEAVE(w)],
+    1 << (PCI_CXL_RANGE_TIMEOUT(w) * 2));
+}
+
+static void
 dvsec_cxl_device(struct device *d, int where, int rev)
 {
+  u32 cache_size, cache_unit_size, l;
+  u64 range_base, range_size;
   u16 w;
 
   /* Legacy 1.1 revs aren't handled */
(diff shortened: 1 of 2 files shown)


ls-ecaps.c
```diff
@@ -701,7 +701,7 @@ cap_dvsec_cxl(struct device *d, int id, int where)
   if (id != 0)
     return;
 
-  if (!config_fetch(d, where + PCI_CXL_CAP, 12))
+  if (!config_fetch(d, where + PCI_CXL_CAP, 0x38 - 0xa))
```
pali: If you are changing _size_ by its calculation then I would suggest to use new/existing macros instead of harcoding magic numbers like `0x38` or `0xa`. There are:
```c
#define PCI_CXL_CAP		0x0a	/* CXL Capability Register */
#define PCI_CXL_DEV_CAP			0x0a	/* CXL Capability Register */
#define PCI_CXL_DEV_RANGE2_BASE_LO	0x34
```
which seems to be source of those magic numbers.

ls-ecaps.c
```diff
@@ -690,35 +690,49 @@ cap_rcec(struct device *d, int where)
...
-  printf(": CXL\n");
-  if (verbose < 2)
-    return;
-
-  if (id != 0)
-    return;
-
-  if (!config_fetch(d, where + PCI_CXL_CAP, 0x38 - 0xa))
+  /* Legacy 1.1 revs aren't handled */
+  if (rev != 1)
```
pali, line 720: Cannot be revision bigger than 1 in future with backward compatible capabilities? IIRC All PCI capabilities were always defined in backward compatible manner. So check for one specific revision seems to be unexpected.

pali (commented):
Hello! I looked at this new version and below are my two comments:

ls-ecaps.c
jphaws: Yes, you are correct. CXL capabilities are defined to be backwards compatible with future revisions. Checking against a specific revision of capabilities without also checking future revisions prevents forward compatibility with later revisions. I'll push an amended commit addressing this.

ls-ecaps.c
```diff
@@ -690,30 +690,206 @@ cap_rcec(struct device *d, int where)
...
+}
+
+static void
+dvsec_cxl_port(struct device *d, int where, int rev)
+{
+  u16 w, m1, m2;
+  u8 b1, b2;
+
+  if (rev != 0)
+    return;
```
pali, line 788: Should not be CXL port capabilities also forward compatible?

ls-ecaps.c
```diff
@@ -690,30 +690,206 @@ cap_rcec(struct device *d, int where)
...
+static void
+dvsec_cxl_device(struct device *d, int where, int rev)
+{
+  u32 cache_size, cache_unit_size, l;
+  u64 range_base, range_size;
+  u16 w;
+
+  /* Legacy 1.1 revs aren't handled */
+  if (rev < 1)
     return;
```
pali, line 721: How is mapped `int rev` to the revision tuple major.minor? Because this check reject rev==0 but comment say that 1.1. So seems like there is non-trivial mapping between integer and tuple major.minor (if it is really major.minor format).

pali (commented):
Hello! Could you look at following two comments?

ls-ecaps.c
jphaws: From 8.1.3 of the CXL 3.0 specification, the DVSEC Revision ID is mapped as follows: 
`
DVSEC Revision ID 0h represents the structure as defined in the CXL 1.1 specification.
`
`
DVSEC Revision ID 1h represents the structure as defined in the CXL 2.0 specification.
`
`
DVSEC Revision ID 2h represents the structure as defined in the CXL 3.0 specification.
`

ls-ecaps.c
jphaws: Yes, they should be. I'll take look back over revision checks throughout and make sure that the code is forward compatible.

ls-ecaps.c
pali: Thanks! Interesting mapping... and code with comment looks correct.

jphaws:
Addressed additional revision checks that would result in capability decoding not being forward compatible with future revisions. Also rebased separate commit that addressed `get_conf_xxx` into commits where `get_conf_xxx` was not used.

ls-ecaps.c
```diff
@@ -690,30 +690,206 @@ cap_rcec(struct device *d, int where)
...
+}
+
+static void
+dvsec_cxl_port(struct device *d, int where, int rev)
+{
+  u16 w, m1, m2;
+  u8 b1, b2;
+
+  if (rev < 0)
+    return;
```
pali, line 788: It is possible that this condition is truth in some case?
Type of `rev` is signed integer but you fill it via `rev = BITS(get_conf_byte(d, where + 0x6), 0, 4);`
which is always non-negative number and of unsigned type `u8`.

ls-ecaps.c
```diff
@@ -690,30 +690,206 @@ cap_rcec(struct device *d, int where)
...
+  printf("\t\t\tRegisterOffset: %016" PCI_U64_FMT_X "\n", (base_hi << 32ULL) | base_lo << 16);
+}
+
+static void
+dvsec_cxl_register_locator(struct device *d, int where, int len, int rev)
+{
+  int i, j;
+
+  if (rev < 0)
+    return;
```
pali, line 836: Same here.

pali (commented):
Just one suspicious thing:

ls-ecaps.c
jphaws: It seems I was mistaken when I though further revisions had been made to this register and the one below. The method used to fill `rev`, as you mentioned, will never fill to anything less than 0. Since this condition will never be true and they are both currently on revision 0 (section 8.1.8 & section 8.1.5), I'm going to remove these revision checks as well.

pali: Ok, now it looks better!

jphaws:
@gollux is there anything else I can do to help this one along?

gollux:
I merged the patch series, but I had to clean it up a lot. Please see my commit in `master` and check it.

Also, I would appreciate if you could add a configuration dump of some device with CXL to `tests/`, so that I can test the code.

I noticed some of the capabilities are still not fully decoded. I can add it, but I do not want to do it without test data as I do not have access to any real CXL device :)

jphaws:
Just opened #115 to address a bug with your commit to `master`, otherwise it looks good to me. 

I have another series pending to decode the remaining capabilities that I'll rebase and clean up over the next few days before sending it out as well. I need to double check before publishing the configuration dump, but it shouldn't be a problem :)

Issue #87, opened 2022-01-12
lspci: Option -d for tree view is broken


pali:
Option `-d` does not work in tree view.

Test case:
```
$ lspci --version
lspci version 3.7.0
$ lspci -tvnn -d 1002:6660
```
No output. But without `-d` it lists device 1002:6660 correctly:
```
$ lspci -tvnn
-[0000:00]-+-00.0  Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor DRAM Controller [8086:0c04]
           +-01.0-[01]----00.0  Advanced Micro Devices, Inc. [AMD/ATI] Sun XT [Radeon HD 8670A/8670M/8690M / R5 M330 / M430] [1002:6660]
...
```
I tested both v3.7.0 git tag and master branch, same results.

It looks like that older version of lspci was able to show device specified by `-d` also in tree view:

```
$ lspci --version
lspci version 3.5.2
$ lspci -tvnn -d 1002:6660
-+-[0000:01]---00.0  Advanced Micro Devices, Inc. [AMD/ATI] Sun XT [Radeon HD 8670A/8670M/8690M / R5 M330 / M430 / Radeon 520 Mobile] [1002:6660]
 \-[0000:00]-
```

PR #112, opened 2022-10-29
lspci: Fix filter support (-s/-d) for subtrees in tree mode (-t)

pali:
Fixes issue https://github.com/pciutils/pciutils/issues/87 (lspci: Option -d for tree view is broken)


gollux:
Thanks, merged!

PR #115, opened 2022-11-01
cxl: Fix Register Locator DVSEC decoding

jphaws:
Fix decoding of register blocks by introducing offset to position calculation (8.1.9 of CXL 3.0 spec) and removed unused defines for Register Locator DVSEC. According to Figure 8-6 (8.1.9 of CXL 3.0 spec), Register Blocks are also indexed starting at 1. 

Signed-off-by: Jaxon Haws <jaxon.haws@amd.com>


gollux:
Thanks, I will merge this.

Could you please contribute a test case (a configuration space dump of a device which has CXL capabilities)?

PR #127, opened 2022-12-22
remove-which-dependency-from-update-pciids.sh

notorand-it:
The builtin `which` doesn't exist in Bash, thus making this script unusable.
The solution is to call the tool itself with some reasonable (and tested) argument.
You get `127` as a return code from a "program not found" situation.

### DIFF
--- a/update-pciids.sh
+++ b/update-pciids.sh
@@ -20,23 +20,23 @@ if [ "$PCI_COMPRESSED_IDS" = 1 ] ; then
 	DECOMP="cat"
 	SRC="$SRC.gz"
 	GREP=zgrep
-elif which bzip2 >/dev/null 2>&1 ; then
+elif bzip2 --help >/dev/null 2>&1 ; then
 	DECOMP="bzip2 -d"
 	SRC="$SRC.bz2"
-elif which gzip >/dev/null 2>&1 ; then
+elif gzip --help >/dev/null 2>&1 ; then
 	DECOMP="gzip -d"
 	SRC="$SRC.gz"
 else
 	DECOMP="cat"
 fi
 
-if which curl >/dev/null 2>&1 ; then
+if curl --version >/dev/null 2>&1 ; then
 	DL="curl -o $DEST.new $SRC"
 	${quiet} && DL="$DL -s -S"
-elif which wget >/dev/null 2>&1 ; then
+elif wget --version >/dev/null 2>&1 ; then
 	DL="wget --no-timestamping -O $DEST.new $SRC"
 	${quiet} && DL="$DL -q"
-elif which lynx >/dev/null 2>&1 ; then
+elif lynx --version >/dev/null 2>&1 ; then
 	DL="eval lynx -source $SRC >$DEST.new"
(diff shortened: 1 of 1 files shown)


update-pciids.sh
```diff
@@ -20,23 +20,23 @@ if [ "$PCI_COMPRESSED_IDS" = 1 ] ; then
 	DECOMP="cat"
 	SRC="$SRC.gz"
 	GREP=zgrep
-elif which bzip2 >/dev/null 2>&1 ; then
+elif bzip2 --help >/dev/null 2>&1 ; then
```
guillemj, line 23: Using `command -v` is better and more portable

guillemj:
Also see #64.

update-pciids.sh
notorand-it: > Using `command -v` is better and more portable

There is no portable way, IMHO, because there is no standard about returns codes and informational arguments.
Instead, I simply checked (by actual testing) which option gives back a 0 return code.

guillemj: POSIX specifies `command -v` (see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html), which any POSIX shell must support (including bash).

notorand-it: >  any POSIX shell must support (including bash)

Uh? I am not talking about the shell, as I haven't changed anything about the shell.
I am talking about those tools used by the script.
`/bin/sh` is a symlink to `bash` in a lot of Linux distrubutions that use `bash` as the default shell.
And `bash` has no `which`  built-in.
That said, let's get back at my PR.

```
$ bzip2 -v; echo $?
bzip2: I won't write compressed data to a terminal.
bzip2: For help, type: `bzip2 --help'.
1
```
`bzip2` is not POSIX-compliant as it doesn't support `-v`.

```
$ curl -v; echo $?
curl: no URL specified!
curl: try 'curl --help' for more information
2
```
`curl` is not POSIX-compliant as it doesn't support `-v`.
```
$ wget -v; echo $?
wget: missing URL
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
1
```
`wget` is not POSIX-compliant as it doesn't support `-v`.
I could continue...

The only portable way I know of to check whether a binary exists is to try to run it with some no-op parameter inside an `if ... then` statement (so it won't break the script in case you have something like `set -eo pipefail`).
If the binary cannot be found, the shell will return usually a 127 return code (which is a -1 in 8 bit) which translated to `false` inside an `if ... then` statement.

notorand-it: > I guess this is where the confusion stems from, `command` is an actual POSIX shell built-in, which supports a `-v` option to lookup any program name and return the pathname found. If you check the commit in the other PR you'll see this is exactly what it's being done there, replacing the literal string `which` with `command -v`. And as mentioned, and referred with the link to POSIX, this is specified and supported as a must by any POSIX conforming shell.

Fair.  It makes sense.

gollux:
Duplicate of #64, which I just closed.

PR #113, opened 2022-10-30
libpci: Add new windows kldbgdrv.sys implementation

pali:
Microsoft Kernel Local Debugging Driver (kldbgdrv.sys) allow access for userspace processes to the PCI config space. It supports access up to 655536 domains and whole 4096 bytes long extended PCIe config space. Driver is signed by Microsoft and is available for both 32-bit and 64-bit systems.

Driver is not part of Windows system and has to be installed via WinDbg installation package. Standalone installers for WinDbg 6.12.2.633 version:
* https://download.microsoft.com/download/A/6/A/A6AC035D-DA3F-4F0C-ADA4-37C8E5D34E3D/setup/WinSDKDebuggingTools_amd64/dbg_amd64.msi
* https://download.microsoft.com/download/A/6/A/A6AC035D-DA3F-4F0C-ADA4-37C8E5D34E3D/setup/WinSDKDebuggingTools/dbg_x86.msi

This kldbgdrv.sys API is used by the !pci command of new WinDbg kernel debugger for displaying PCI config space.

API of this driver is available only for processes with Debug privilege and only if system was booted with Debugging option.

PR #129, opened 2022-12-30
libpci: win32-cfgmgr32: Add support for accessing config space via other backend

pali:
Extend win32-cfgmgr32 backend and add a new option win32.cfgmethod for specifying other backend for accessing PCI config space. There are more config space access methods available on Windows and each is working only sometimes (either requires special privileges or special setup).

So by default try to choose the first working one via order defined in pci probe_sequence[] array. If none is available then emulate PCI config space like before this change.

Function pci_init_v35() is extended and renamed to pci_init_internal() to optionally do not throw errors and allow to specify one access method which will be skipped in AUTO mode. This is used to prevent choosing win32-cfgmgr32 as config space access method for win32-cfgmgr32.

### DIFF
--- a/pcilib.man
+++ b/pcilib.man
@@ -86,8 +86,12 @@ Device listing on Windows systems using the Windows Configuration Manager
 via cfgmgr32.dll system library. This method does not require any special
 Administrator rights or privileges. Configuration Manager provides only basic
 information about devices, assigned resources and device tree structure. There
-is no access to the PCI configuration space but libpci provides read-only
-virtual emulation based on information from Configuration Manager. Starting
+is no access to the PCI configuration space but libpci either tries to use
+other access method to access configuration space or it provides read-only
+virtual emulation based on information from Configuration Manager. Other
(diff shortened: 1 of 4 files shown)


pali:
All issues from comment https://github.com/pciutils/pciutils/pull/113#issuecomment-1296275984 should be addressed now.

pcilib.man
```diff
@@ -165,6 +169,15 @@ Physical addresses of memory-mapped I/O ports for Extended PCIe Intel configurat
 It has same format as
 .B mmio-conf1.addrs
 parameter.
+.TP
+.B win32.cfgmethod
+Config space access method for win32-cfgmgr32 on Windows systems. Value
+.I auto
+or emtpy string probe and choose the first access method which supports access
```
gollux, line 176: Typo `emtpy`.

gollux:
I like the new version much more, but I still do not agree with the way it handles errors. I will fix it after merging.

PR #130, opened 2023-01-03
debugpci: Add tool for diagnosing common PCIe errors

shradhat:
This tool captures and dumps common PCIe link errors for PCIe devices on the bus which use the Synopsys DesignWare controllers.
The DW controllers provide software mechanisms to detect PCIe link errors. Capturing and dumping this debug information in a systematic method helps reduce effort in debugging common PCIe issues. It also provides debug suggestions for the different errors wherever possible

Signed-off-by: Rishabh Dash <rishabh.dash3601@gmail.com>
Signed-off-by: Shradha Todi <shradha.t@samsung.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ PCIINC_INS=lib/config.h lib/header.h lib/pci.h lib/types.h
 
 export
 
-all: lib/$(PCIIMPLIB) lspci$(EXEEXT) setpci$(EXEEXT) example$(EXEEXT) lspci.8 setpci.8 pcilib.7 pci.ids.5 update-pciids update-pciids.8 $(PCI_IDS)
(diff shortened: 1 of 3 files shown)


gollux:
Sorry, but this is a too much specialized tool to be installed by default. Could you please move it to a `contrib` directory and build it only when requested?

pali:
This tool is parsing Synopsys Designware PCI Bridge / PCIe Root Port capabilities and printing them in human readable format, right? What about rather extending existing https://github.com/pciutils/pciutils/blob/master/ls-caps-vendor.c file and use existing lspci tool?

gollux:
I agree with @pali that the better way to go is to add it to `lspci` as a vendor-specific capability.

Could you please try it?

shradhat:
I will work to implement this as an extension of lspci. If I face any hiccup, will update. Thanks for your suggestions!

PR #135, opened 2023-03-05
i386-ports: Remove cpu endianity swapping

pali:
All cpu_to_le*() and le*_to_cpu() functions on x86 are identity functions because x86 is little endian. Moreover config space buffers in libpci are stored in cpu endianity therefore conversion to little endian is wrong.


gollux:
In personal discussion, we concluded that the current behavior is actually correct and the only thing lacking are comments explaining that all representations of the config space as a byte sequence in libpci are always little-endian. I will add them.

PR #144, opened 2023-05-07
Fix -flto=thin compiling issue

### LINKED ISSUE #143: pciutils@0766bfa0f557430a6dd3256d45a5b756499f9f7b fails to build with clang -flto=thin
Building pciutils at commit 0766bfa0f557430a6dd3256d45a5b756499f9f7b fails when using the following command:

```
make -j8
    HOST=x86_64-pc-linux-gnu                                                                \
    CROSS_COMPILE=x86_64-pc-linux-gnu-                                                      \
    'CC=clang -mfpmath=sse'                                                                 \
    AR=llvm-ar                                                                              \
    PKG_CONFIG=x86_64-pc-linux-gnu-pkg-config                                               \
    RANLIB=llvm-ranlib                                                                      \
    DNS=no                                                                                  \
    IDSDIR='$(SHAREDIR)/hwdata'                                                             \
    MANDIR='$(SHAREDIR)/man'                                                                \
    PREFIX=/usr                                                                             \
    SHARED=yes                                                                              \
    STRIP=ZLIB=yes                                                                          \
    PCI_COMPRESSED_IDS=0                                                                    \
    PCI_IDS=pci.ids                                                                         \
    LIBDIR=${PREFIX}/lib                                                                    \
    LIBKMOD=no                                                                              \
    HWDB=yes                                                                                \
    'OPT=-march=native -O2 -pipe -D_FORTIFY_SOURCE=3 -fstack-protector-strong -flto=thin'   \
    all
```

The errors:
```
ld.lld: error: filter.o <inline asm>:13:1: default version symbol pci_fill_info@@LIBPCI_3.8 must be defined
.symver pci_fill_info_v38, pci_fill_info@@LIBPCI_3.8
^


ld.lld: error: generic.o <inline asm>:1:1: default version symbol pci_fill_info@@LIBPCI_3.8 must be defined
.symver pci_fill_info_v38, pci_fill_info@@LIBPCI_3.8
^


ld.lld: error: caps.o <inline asm>:1:1: default version symbol pci_fill_info@@LIBPCI_3.8 must be defined
.symver pci_fill_info_v38, pci_fill_info@@LIBPCI_3.8
```

The expected result is a successful compile

The full build:
[build.log](https://github.com/pciutils/pciutils/files/11415155/build.log)

Additional details:
[emerge_info.log](https://github.com/pciutils/pciutils/files/11415159/emerge_info.log)

```
% clang --version
clang version 16.0.2
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm/16/bin
Configuration file: /etc/clang/clang.cfg
```

nvinson:
Fix based on recommendations from
https://maskray.me/blog/2020-11-26-all-about-symbol-versioning#recommended-usage

Fixes #143

### DIFF
--- a/lib/header.h
+++ b/lib/header.h
@@ -254,6 +254,7 @@
 #define PCI_EXT_CAP_ID_LMR	0x27	/* Lane Margining at Receiver */
 #define PCI_EXT_CAP_ID_HIER_ID	0x28	/* Hierarchy ID */
 #define PCI_EXT_CAP_ID_NPEM	0x29	/* Native PCIe Enclosure Management */
+#define PCI_EXT_CAP_ID_32GT	0x2a	/* Physical Layer 32.0 GT/s */
 #define PCI_EXT_CAP_ID_DOE	0x2e	/* Data Object Exchange */
 
 /*** Definitions of capabilities ***/
--- a/ls-ecaps.c
+++ b/ls-ecaps.c
@@ -1556,6 +1556,9 @@ show_ext_caps(struct device *d, int type)
 	  case PCI_EXT_CAP_ID_NPEM:
 	    printf("Native PCIe Enclosure Management <?>\n");
 	    break;
+      case PCI_EXT_CAP_ID_32GT:
+	    printf("Physical Layer 32.0 GT/s <?>\n");
+        break;
 	  case PCI_EXT_CAP_ID_DOE:
 	    cap_doe(d, where);
 	    break;


pali:
This change is breaking compilation for windows target by gcc compiler.

For example all function calls to symbol pci_fill_info are being resolved to function pci_fill_info() instead of the correct one pci_fill_info_v38().
```
/usr/bin/i686-w64-mingw32-ld: warning: resolving _pci_fill_info@@LIBPCI_3.8 by linking to _pci_fill_info
```

And in shared build, because symbol pci_fill_info@@LIBPCI_3.8 does not exist anymore with this change, GNU LD resolves it to _pci_fill_info symbol (which is the older version).
```
/usr/bin/i686-w64-mingw32-ld: warning: resolving _pci_fill_info by linking to _pci_fill_info@@@LIBPCI_3.8
Use --enable-stdcall-fixup to disable these warnings
Use --disable-stdcall-fixup to disable these fixups
```

This issue was fixed in commit 26a64f917656453051d200cc7bf87710a0bfcef7 but this change is reverting it.

nvinson:
>

nvinson:
Sorry for the noise. I accidentally hit the wrong button.

pali:
I'm really not sure if usage of @@@ instead of @@ on any non-win32 platform is correct.

But you can avoid that win32 ifdef by modifying lib/ver2def.pl script to generate build symbol with 3 @@@ characters (instead of 2) and then GNU LD should hopefully handle new symbols with @@@.

nvinson:
> I'm really not sure if usage of @@@ instead of @@ on any non-win32 platform is correct.

The usage of '@@@' is defined in MaskRay's blog (link in commit message). It's also documented in GNU assembler's manual (https://sourceware.org/binutils/docs/as/Symver.html). The symver directive only works for ELF platforms.

> 
> But you can avoid that win32 ifdef by modifying lib/ver2def.pl script to generate build symbol with 3 @@@ characters (instead of 2) and then GNU LD should hopefully handle new symbols with @@@.

I can make that change, but I don't have a mingw compiler or the clang equivalent. I won't be able to test the change.

nvinson:
> I can make that change, but I don't have a mingw compiler or the clang equivalent. I won't be able to test the change.

I don't believe the modifying the perl script is necessary. After a few more reviews, I don't think the ifdefs are needed either. I've removed them.

pali:
Your last change still breaks Windows version compiled by GCC. Now it compiles but trying to start lspci.exe fails with error:
The procedure entry point `pci_filter_parse_slot@@LIBPCI_3.8` could not be loacated in the dynamic library `libpci3.dll`.

It is because `lspci.exe` is linked with `pci_filter_parse_slot@LIBPCI_3.8` symbol and symbol `pci_filter_parse_slot@LIBPCI_3.8` in `libpci3.dll` is forwareded to symbol `pci_filter_parse_slot@@LIBPCI_3.8` which you removed in this change.

Basically with this change there is no implementation for version 3.8 of function pci_filter_parse_slot() in the final libpci3.dll library as GNU LD is being instructed to not include it.

pali:
Still does not work with GNU LD, now it is failing on error:
The procedure entry point `pci_init@@@LIBPCI_3.5` could not be located in the dynamic link library `libpci3.dll`.

I have feeling that clang and GNU LD needs different handling of the versioned symbols because current version is working with GNU LD. Also there were commits for fixing GNU LD: f514df26e4ad34755607b392af8ef2cb81bb296d 0478e1f3928bfaa34eb910ba2cbaf1dda8f84aab

So maybe the correct way is to introduce some ifdef clang and put there clang specific versioning code?

nvinson:
> Still does not work with GNU LD, now it is failing on error: The procedure entry point `pci_init@@@LIBPCI_3.5` could not be located in the dynamic link library `libpci3.dll`.
> 
> I have feeling that clang and GNU LD needs different handling of the versioned symbols because current version is working with GNU LD. Also there were commits for fixing GNU LD: [f514df2](https://github.com/pciutils/pciutils/commit/f514df26e4ad34755607b392af8ef2cb81bb296d) [0478e1f](https://github.com/pciutils/pciutils/commit/0478e1f3928bfaa34eb910ba2cbaf1dda8f84aab)
> 
> So maybe the correct way is to introduce some ifdef clang and put there clang specific versioning code?

The issue was because I didn't update the SYMBOL_VERSION macro for pci_init().  This update fixes that oversight.

With help from [ionen](https://wiki.gentoo.org/wiki/User:Ionen), I was able to test lspci with wine. It doesn't produce a successful result, but it did print `lspci: Cannot find any working access method.` instead of an error about missing symbols, so I think lspci.exe is running correctly.

pali:
Now Windows version compiled by GNU LD works.

Note that wine is silly, it does not implement most of the cfgmgr32 API and returns success for all unimplemented operation which should allocate and fill buffers. So PCI or HW access in lspci via wine would not work. But local file access via `-F` should work.

PR #146, opened 2023-06-05
CXL: Flex Bus cap; Test Capability (0xA)

alexisgrytalms:
* Fix the Flex Bus DVSEC capability 
* Add some checks on the length
* Add the PCIe DVSEC for Test Capability `0xA`
* Print the name of the CXL DVSEC capability


ls-ecaps.c
```diff
@@ -950,15 +950,33 @@ dvsec_cxl_gpf_port(struct device *d, int where)
...
+      break;
+    case 2:
+      if (len!=0x20) {
+        printf("\t\t<Wrong length for Revision %d>\n", rev);
+        return;
+      }
+      break;
+    default:
+      printf("\t\tRevision %d not supported\n", rev);
+      return;
```
pali, line 979: Revisions should be forward-compatible, right? Then this code disallow forward compatibility with future rev3.

I think that instead of checking length of data at this place at the beginning of the function, length should be checked later, when trying to access data at requested offset (that length is larger than accessing offset).

alexisgrytalms: I fixed it.

ls-ecaps.c
```diff
@@ -950,17 +950,33 @@ dvsec_cxl_gpf_port(struct device *d, int where)
...
   u32 l, data;
 
-  if (rev < 1)
-  {
-    printf("\t\tRevision %d not supported\n", rev);
-    return;
+  // Sanity check: Does the length correspond to its revision?
+  switch (rev) {
+    case 0:
+      if (len != 0x10) {
```
pali, line 961: Length is the offset of the first unsupported register (first register after the last supported) in specified version, right? So what about using named register constant instead of magic number? E.g.

```suggestion
      if (len != PCI_CXL_FB_MOD_TS_DATA) {
```

ls-ecaps.c
```diff
@@ -993,12 +1009,18 @@ dvsec_cxl_flex_bus(struct device *d, int where, int rev)
...
-  l = get_conf_long(d, where + PCI_CXL_FB_MOD_TS_DATA);
-  data = BITS(l, 0, 24);
-  printf("\n\t\tFBModTS:\tReceived FB Data: %06x\n", (unsigned int)data);
+  // From Rev 1
+  if (rev >= 1)
+  {
+    l = get_conf_long(d, where + PCI_CXL_FB_MOD_TS_DATA);
+    data = BITS(l, 0, 24);
+    printf("\t\tFBModTS:\tReceived FB Data: %06x\n", (unsigned int)data);
+  }
```
pali, line 1020: You are missing a check here that register `PCI_CXL_FB_MOD_TS_DATA` is really available, so to avoid reading data from uninitialized/unavailable config space. In previous suggest I wrote that for this it could make sense to move length check from the beginning of the function to this place, and call code for reading parsing this register conditionally. E.g.:

```suggestion
  if (rev >= 1 && len >= PCI_CXL_FB_MOD_TS_DATA + 4)
  {
    l = get_conf_long(d, where + PCI_CXL_FB_MOD_TS_DATA);
    data = BITS(l, 0, 24);
    printf("\t\tFBModTS:\tReceived FB Data: %06x\n", (unsigned int)data);
  }
```

lib/header.h
```diff
@@ -1069,7 +1069,7 @@
 #define PCI_DVSEC_ID_CXL	0	/* Designated Vendor-Specific ID for Intel CXL */
 
 /* PCIe CXL Designated Vendor-Specific Capabilities for Devices: Control, Status */
-#define PCI_CXL_DEV_LEN 0x38 /* CXL Device DVSEC Length */
+#define PCI_CXL_DEV_LEN 0x3c /* CXL Device DVSEC Length */
```
pali, line 1072: This change looks suspicious. Either PCI_CXL_DEV_LEN is constant and should have its _constant_ value - which is not changing - and have its backward compatible ABI. Or it is variable (e.g. based on the current supported version in the lspci) and then it is not constant at all and should not be in header file defined as a compile time constant for all programs which use libpci API/ABI.

alexisgrytalms: The length is dependent on the revision, therefore isn't constant but always increasing.
Same reason, I followed the existing code from MR #120 that was accepted.

I considered `PCI_CXL_DEV_LEN` as the maximum length the current lspci version supports.
The actual length of the structure exposed by the device is read from the DVSEC header.

It's fine to keep it, it shows up to what length lspci supports the structure.

PR #152, opened 2023-07-18
lspci: Add support for the DOE Discovery options

alistair23:
The PCIe 6 specification added support for the Data Object Exchange (DOE). When DOE is supported the Discovery Data Object Protocol must be implemented. The protocol allows a requester to obtain information about the other DOE protocols supported by the device.

This patch adds support for querying the list of DOE Discovery Data Object Protocols supported by the device.

This unfortunately does require writing to the device, as we can't query the information without sending data objects to the device.

### DIFF
--- a/lib/header.h
+++ b/lib/header.h
@@ -1393,6 +1393,12 @@
 #define  PCI_DOE_STS_INT		0x2	/* DOE Interrupt Status */
 #define  PCI_DOE_STS_ERROR		0x3	/* DOE Error */
 #define  PCI_DOE_STS_OBJECT_READY	0x80000000 /* Data Object Ready */
+#define PCI_DOE_WRITE_MAILBOX		0x10	/* DOE Write Data Mailbox Register */
+#define PCI_DOE_READ_MAILBOX		0x14	/* DOE Read Data Mailbox Register */
+
+#define PCI_DOE_DISCOVERY_RESPONSE_VENDOR_ID	0xFFFF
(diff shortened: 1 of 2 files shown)


jic23:
Hi Alistair,

Whilst I'd love to see this... It is not safe to do unmediated accesses to the DOE if there is any potential for a driver being bound to it, (or shortly in Linux at least if the PCI core is using it).  You can end up reading and clearing the response to someone else's request - and drivers tend not to assume someone will steal their data from under them.  That's particularly 'fun' if you break a security related transfer and end up taking down the device due to a key refresh or similar failing.

From linux side of things, we decided early on that any access must be mediated by the PCI driver stack.  I'd imagine other operating systems will do the same.  So this path is a non starter as far as I'm concerned. 

For Linux, we could add some sysfs files to provide this information given it will have been queried anyway on all PCI devices that have DOEs (after Lukas finishes the CMA  / attestation support).  lspci could read that.  If you have a specific protocol that is of interest to you then we should look at a proper safe userspace interface.

gollux:
I completely agree that this should not be done by default. In particular, `lspci` without special options should never write to configuration registers.

I can imagine implementing this as a separate tool for debugging hardware, but I believe that the right place is in the kernel.

alistair23:
I suspected writing directly to the device wasn't a viable upstream solution.

I wanted to get some feedback about if this is something we wanted to do in userspace via special arguments or it it had to be in the kernel. It sounds like the kernel is the place to do this and then we can query the information from there. I'll work towards that

PR #153, opened 2023-07-19
lib: fixup DOE status register bit

twilfredo:
The error bit is specified by the 2nd (zero indexed) bit in the status register, so the respective bit value is 4 (PCI Base Spec 6.0.1). Let's fix that up.


gollux:
Thanks, merged.

PR #154, opened 2023-07-19
RFC: lspci: Use mangled vendor/device ID when examining vendor caps

dme:
Given that PCI VFs are expected to have a vendor and device ID of 0xffff, when examining vendor capabilities use the mangled vendor and device IDs (typically copied from the PF) rather than those read from the VF configuration space.


pali (approved):
Looks good!

gollux:
Thanks, merged!

PR #155, opened 2023-09-01
Add support for 32.0 GT/s header

LeonSkoog:
PCie Gen5 becoming more and more popular.
Let's add header parser for  'Physical Layer 32.0 GT/s Extended Capability Header'


gollux:
Thanks!

pali:
Btw @LeonSkoog you have there some codying style inconsistency...

And have thinking about writing a change to fully decode this extended capability?
Because there is just a `<?>` in the output.

Issue #21, opened 2018-12-18
Perhaps use the meson buildsystem


hughsie:
I'm building libpci as part of a flatpak, which has to be built on various obscure architectures. With the custom Makefile-based build system it's quite hard to set up the various different prefixes and build options so that libpci is usable by other libraries and binaries such as flashrom.

Rather than add various kludges to the Makefiles (which I initially did), it was actually easier and more understandable for me to just add support for the meson buildsystem. Meson is a fairly new project, but it already has a lot of users including core projects like wayland, the x-server and tons of GUI applications. I'm not a meson developer or advocate, just a content user who thinks it's quite a nice project.

I know hacking on build systems isn't useful time fixing bugs, but I do think pciutils would benefit from making everything simpler. Meson support adds just 147 lines for similar functionality to Make (i.e. pkgconfig files, header includes, shared library, config arguments, 2xCLI tools), although admittedly I've only added the Linux bits. Adding support for other OS's would probably add another 50 lines to the PR.

That said, I don't want to spent any more time on this than required; if you said "I'm quite happy with a Makefile" then I can just close this issue, use my patch to built flashrom as a flatpak and then move onto other things. If you say that the project might has outgrown the Makefile as a build system then I'm happy to do the extra work and add the `if`s for the various BSDs and OS-X etc. I've noticed the Fedora distro package for pciutils basically re-implements the install functionality, including properly installing the shared library on Linux. It seems other distros do similar.

Let me know what you think. If you're curious I can submit a PR but didn't want to be overly pushy. Thanks.

gollux:
I would very much like to avoid dependency on Meson. However, adding an alternative build system could be an interesting experiment. So I am definitely interested in a patch.

eli-schwartz:
> I would very much like to avoid dependency on Meson.

FWIW, if your concern is dragging in a python dependency: https://mesonbuild.com/FAQ.html#but-i-really-want-a-version-of-meson-that-doesnt-use-python

This is really a bit of a recent development, but.

There is a second implementation of meson called "muon" that is very simple to bootstrap with just a shell and a c99 compiler. No python required. While indeed that is a new dependency, it's not a hard one (at least if your motivation is about bootstrapping concerns) and in return it does let you drop a dependency on gmake (I notice the current Makefile is NOT using the portable subset of Make syntax).

Of course, supporting both is also a valid alternative.

nicola-lunghi:
[...]

PR #156, opened 2023-09-09
Example meson build system

nvinson:
This is a mostly-complete example of a meson build system from pciutils. The windows piece needs to be finished and the non-Linux platforms would need testing.

Unfortunately, I don't have the resources or knowledge to complete those pieces.


nvinson:
This addresses the issue raised in #21

pali:
I see that new meson build script calls shell commands. Cannot be this avoided?

Also Windows building is missing there.

pali:
I have tried to build pciutils from this PR on one x86 Linux machine, but it failed:
```
The Meson build system
Version: 0.49.2
Source dir: /tmp/pciutils
Build dir: /tmp/pciutils/build
Build type: native build

meson.build:1:0: ERROR:  Meson version is 0.49.2 but project requires >= 1.2.0.

A full log can be found at /tmp/pciutils/build/meson-logs/meson-log.txt
```
(repository does not contain any newer meson version)


Next I have tried it on other ppc box, but it failed too. Repository has only meson 0.54.

pali:
I downloaded meson directory from https://github.com/mesonbuild/meson to ppc box but it still does not want to compile pciutils...

```
The Meson build system
Version: 1.2.99
Source dir: /tmp/pciutils
Build dir: /tmp/pciutils/build
Build type: native build
Project name: pciutils
Project version: 3.10.0
C compiler for the host machine: gcc (gcc 7.4.0 "gcc (OpenWrt GCC 7.4.0) 7.4.0")
C linker for the host machine: gcc ld.bfd 2.27
Host machine cpu family: ppc
Host machine cpu: ppc
Did not find pkg-config by name 'pkg-config'
Found pkg-config: NO
Did not find CMake 'cmake'
Found CMake: NO
Run-time dependency zlib found: NO (tried pkgconfig, cmake and system)
Run-time dependency libkmod found: NO (tried pkgconfig and cmake)
Run-time dependency libudev found: NO (tried pkgconfig and cmake)
Has header "resolv.h" : YES
Library resolv found: NO
Checking for function "res_query" with dependency -lresolv: YES
Configuring config.h using configuration

../meson.build:277:0: ERROR: Program 'ctags' not found or not executable

A full log can be found at /tmp/pciutils/build/meson-logs/meson-log.txt
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```

And ctags package is not available in repository. Is there reason why ctags is required for meson, but not for make?

nvinson:
> I see that new meson build script calls shell commands. Cannot be this avoided?
> 
> Also Windows building is missing there.

Not at all surprising since I stated "The windows piece needs to be finished..." in an earlier comment.



> (repository does not contain any newer meson version)
> 
> Next I have tried it on other ppc box, but it failed too. Repository has only meson 0.54.

What repository are you referring to?
The latest version of meson is 1.2.1 and is available at https://github.com/mesonbuild/meson/releases

> And ctags package is not available in repository. Is there reason why ctags is required for meson, but not for make?

ctags is required for the `make ctags` rule. That said, the `meson.build` script was not supposed to require `ctags` for standard building. I will make updates to `meson.build` to remove this requirement.

> I see that new meson build script calls shell commands. Cannot be this avoided?

`Makefile` uses the same shell commands (with the exception of `date`); therefore, before I go through the effort of removing all shell commands, I would an assurance that the pciutils project is willing to move forward with meson.

Moreover, the dependency on `date` is  optional, and could be removed completely.

meson.build
```diff
@@ -0,0 +1,519 @@
...
+                 gnu_symbol_visibility: 'hidden',
+                 link_args: '-Wl,--version-script=' + '../lib/libpci.ver',
+                 link_depends: ['lib/libpci.ver'])
+
+pkg = import('pkgconfig')
+pkg.generate(pcilib,
+             name: 'libpci',
+             filebase: 'libpci',
+             description:'libpci',
+             variables: {'idsdir': IDSDIR})
```
eli-schwartz, line 257: This can be changed to:
```suggestion
             variables: ['idsdir= +'IDSDIR])
```

further dropping the minimum meson version to 0.53.0

meson.build
```diff
@@ -0,0 +1,519 @@
...
+  pci_ids = 'pci.ids.gz'
+else
+  conf.set_quoted('PCI_IDS', 'pci.ids')
+  pci_ids = 'pci.ids'
+endif
+
+if (get_option('with-dns').enabled() or
+  get_option('with-dns').auto() and host_machine.system() != 'windows')
+
+  assert(cc.has_function('res_query', dependencies:dns_dep))
```
eli-schwartz, line 232:
```suggestion
  assert(cc.has_function('res_query', dependencies:dns_dep), 'missing required function: res_query')
```

will drop the minimum required version from 0.53.0 to 0.50.0

eli-schwartz (commented):
Thanks for removing the `preserve_path: DEFAULT_VALUE` that was totally extraneous -- and the easy `mv  meson.options meson_options.txt` lowering the requirements quite a bit.

There's a couple more easy changes you could make too.

nvinson:
Made a few updates to `meson.build` and renamed `meson.options` to `meson_options.txt`. These changes let met set the minimum required meson version to 0.53.0

eli-schwartz:
> ctags is required for the `make ctags` rule. That said, the `meson.build` script was not supposed to require `ctags` for standard building. I will make updates to `meson.build` to remove this requirement.

Incidentally, meson will automatically create a "ctags" target for you if you didn't define one yourself *and* ctags is installed. It will chdir to the project source root and use git ls-tree to determine which files to run on. Dunno how badly one might want to control the precise list of files, but in theory you can get away with just skipping that whole step. Or preserve your own more-specific target, if you like. :)

meson.build
```diff
@@ -0,0 +1,521 @@
...
+                's#@IDSDIR@#' + IDSDIR + '#',
+                '-e',
+                's#@PCI_IDS@#' + pci_ids + '#',
+                '@INPUT@'
+              ],
+              capture: true,
+              input: [ 'pci.ids.man' ],
+              output: [ 'pci.ids.5' ],
+              install: true,
+              install_dir: get_option('mandir') / 'man5')
```
eli-schwartz, line 440: You can get rid of all these seds using a meson builtin approach to configuring files:

```meson
replacements = {
  'TODAY': build_date,
  'VERSION': 'pciutils-'+meson.project_version(),
  'IDSDIR': IDSDIR,
  'PCI_IDS': pci_ids,
}

foreach man: ['pci.ids.5', 'pcilib.7', 'lspci.8', 'setpci.8', 'update-pciids.8']
    manfile = configure_file(
        input: man.substring(0, -1)+'man',
        output: man,
        configuration: replacements
    )
    install_man(manfile)
endforeach
```

meson.build
```diff
@@ -0,0 +1,521 @@
...
+sources += ['lib/dump.c']
+
+conf_file = configure_file(configuration: conf, output: 'config.h')
+
+pcilib = library('pci',
+                 dependencies: dependencies,
+                 install: true,
+                 sources: sources,
+                 gnu_symbol_visibility: 'hidden',
+                 link_args: '-Wl,--version-script=' + '../lib/libpci.ver',
```
eli-schwartz, line 249:
```suggestion
                 link_args: '-Wl,--version-script=' + meson.current_source_dir() +'/lib/libpci.ver',
```

pali:
> > I see that new meson build script calls shell commands. Cannot be this avoided?
> > Also Windows building is missing there.
> 
> Not at all surprising since I stated "The windows piece needs to be finished..." in an earlier comment.

Ok.

> > (repository does not contain any newer meson version)
> > Next I have tried it on other ppc box, but it failed too. Repository has only meson 0.54.
> 
> What repository are you referring to? The latest version of meson is 1.2.1 and is available at https://github.com/mesonbuild/meson/releases

I'm referring to the repository of system packages available in the on that ppc box, which can be additionally installed via system packaging tool.

> > And ctags package is not available in repository. Is there reason why ctags is required for meson, but not for make?
> 
> ctags is required for the `make ctags` rule. That said, the `meson.build` script was not supposed to require `ctags` for standard building. I will make updates to `meson.build` to remove this requirement.

Thanks!

> > I see that new meson build script calls shell commands. Cannot be this avoided?
> 
> `Makefile` uses the same shell commands (with the exception of `date`); therefore, before I go through the effort of removing all shell commands, I would an assurance that the pciutils project is willing to move forward with meson.
> 
> Moreover, the dependency on `date` is optional, and could be removed completely.

Ok. Makes sense.

> > ctags is required for the `make ctags` rule. That said, the `meson.build` script was not supposed to require `ctags` for standard building. I will make updates to `meson.build` to remove this requirement.
> 
> Incidentally, meson will automatically create a "ctags" target for you if you didn't define one yourself _and_ ctags is installed. It will chdir to the project source root and use git ls-tree to determine which files to run on. Dunno how badly one might want to control the precise list of files, but in theory you can get away with just skipping that whole step. Or preserve your own more-specific target, if you like. :)

Can we avoid doing any such automagic actions? For building whole pciutils project there is no need for ctags. And I highly doubt that end users or developers need it have enabled by default (and independently of the fact if some binary named ctags is in $PATH or not).

pali:
I have tried to build it again on ppc box with updates from this PR, but it again failed:
```
The Meson build system
Version: 1.2.99
Source dir: /tmp/pciutils
Build dir: /tmp/pciutils/build
Build type: native build
Project name: pciutils
Project version: 3.10.0
C compiler for the host machine: gcc (gcc 7.4.0 "gcc (OpenWrt GCC 7.4.0) 7.4.0")
C linker for the host machine: gcc ld.bfd 2.27
Host machine cpu family: ppc
Host machine cpu: ppc
Did not find pkg-config by name 'pkg-config'
Found pkg-config: NO
Did not find CMake 'cmake'
Found CMake: NO
Run-time dependency zlib found: NO (tried pkgconfig, cmake and system)
Run-time dependency libkmod found: NO (tried pkgconfig and cmake)
Run-time dependency libudev found: NO (tried pkgconfig and cmake)
Has header "resolv.h" : YES
Library resolv found: NO
Checking for function "res_query" with dependency -lresolv: YES
Configuring config.h using configuration

../meson.build:314:0: ERROR: File lib/config.h does not exist.

A full log can be found at /tmp/pciutils/build/meson-logs/meson-log.txt
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```

pali:
> Did not find CMake 'cmake'
> Found CMake: NO

Why is meson searching for cmake? I was expecting that meson does not depend on cmake...

eli-schwartz:
> Can we avoid doing any such automagic actions? For building whole pciutils project there is no need for ctags. And I highly doubt that end users or developers need it have enabled by default (and independently of the fact if some binary named ctags is in $PATH or not).

It is just a "run" target, which means it's never executed unless the user goes out of their way to run it. This is the same behavior as https://github.com/pciutils/pciutils/blob/db43fb5e8f2c04c409bdd06ac2c2828685038d69/Makefile#L134-L136 which is also "enabled" as a Makefile rule by default (but never run unless someone executes `make ctags`).

> Why is meson searching for cmake? I was expecting that meson does not depend on cmake...

When meson looks for the zlib dependency using `dependency('zlib')`, it first tries to run `pkg-config` to find it since that's the standard for software to describe dependencies. However, there is a lot of software that only provides `foobar-config.cmake` files, and meson doesn't really have a good way of knowing which ones those are. So if it cannot be found with pkg-config then meson will default to running a hacky cmake script that tries to find one of those foobar-config.cmake files and parse it using the cmake interpreter and export a debug dump of the variables it sets.

If you know that the dependency is well behaved and provides a pkg-config file, you can disable this by explicitly providing a method:

```meson
dependency('zlib', method: 'pkg-config')
```

(And zlib is one such well-behaved project.)

pali:
Now I have tried the system version of meson and ninja but compilation still fails:
```
$ meson .. && ninja
The Meson build system
Version: 0.54.0
Source dir: /tmp/pciutils
Build dir: /tmp/pciutils/build
Build type: native build
Project name: pciutils
Project version: 3.10.0
C compiler for the host machine: gcc (gcc 7.4.0 "gcc (OpenWrt GCC 7.4.0) 7.4.0")
C linker for the host machine: gcc ld.bfd 2.27
Host machine cpu family: ppc
Host machine cpu: ppc
Did not find pkg-config by name 'pkg-config'
Found Pkg-config: NO
Run-time dependency zlib found: NO (tried pkgconfig)
Run-time dependency libkmod found: NO (tried pkgconfig)
Run-time dependency libudev found: NO (tried pkgconfig)
Has header "resolv.h" : YES
Library resolv found: NO
Checking for function "res_query" with dependency -lresolv: YES
Configuring config.h using configuration
Build targets in project: 10

Found ninja-1.10.2 at /usr/bin/ninja
[5/39] Compiling C object 'lspci@exe/ls-kernel.c.o'
FAILED: lspci@exe/ls-kernel.c.o
gcc -Ilspci@exe -I. -I.. -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -O2 -g -Wno-parentheses -Wstrict-prototypes -Wmissing-prototypes -MD -MQ 'lspci@exe/ls-kernel.c.o' -MF 'lspci@exe/ls-kernel.c.o.d' -o 'lspci@exe/ls-kernel.c.o' -c ../ls-kernel.c
../ls-kernel.c:22:10: fatal error: libkmod.h: No such file or directory
 #include <libkmod.h>
          ^~~~~~~~~~~
compilation terminated.
[7/39] Compiling C object 'lspci@exe/ls-caps.c.o'
ninja: build stopped: subcommand failed.
```

meson.build
```diff
@@ -0,0 +1,481 @@
...
+
+if (get_option('with-dns').enabled() or
+  get_option('with-dns').auto() and host_machine.system() != 'windows')
+
+  assert(cc.has_function('res_query', dependencies:dns_dep), 'missing required function: res_query')
+  conf.set('PCI_USE_DNS', 1)
+  conf.set_quoted('PCI_ID_DOMAIN', 'pci.id.ucw.cz')
+endif
+
+conf.set10('PCI_USE_LIBKMOD', kmod_dep.found())
```
eli-schwartz, line 238: https://github.com/pciutils/pciutils/blob/db43fb5e8f2c04c409bdd06ac2c2828685038d69/ls-kernel.c#L20-L22

ifdef, not if. You can't use set10 here, as this still creates the define unconditionally (but sometimes set to 0), instead, use:
```meson
if kmod_dep.found()
    conf.set('PCI_USE_LIBKMOD', true)
endif
```

meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option('build-date', type: 'string', description: 'build date. Defaults to today if not set')
+option('enable-compat-getopt', type: 'boolean', value: false, description: 'enable getopt compatibility support')
```
pali, line 2: It is possible to autodetect it?

eli-schwartz: https://github.com/pciutils/pciutils/blob/db43fb5e8f2c04c409bdd06ac2c2828685038d69/README.Windows#L30-L31

Meson could autodetect this by compiling a test program and seeing if it works. Under what circumstances is the compat-getopt needed? Is it enough to check if getopt is available in the libc?

pali: It is needed when libc/system does not provide one, mostly for older mingw32 runtime.

meson.build
nvinson: fixed.

meson_options.txt
nvinson: > It is possible to autodetect it?

updated to auto-detect if compatibility getopt is needed.

pali:
Still does not compile via meson:
```
The Meson build system
Version: 0.54.0
Source dir: /tmp/pciutils
Build dir: /tmp/pciutils/build
Build type: native build
Project name: pciutils
Project version: 3.10.0
C compiler for the host machine: gcc (gcc 7.4.0 "gcc (OpenWrt GCC 7.4.0) 7.4.0")
C linker for the host machine: gcc ld.bfd 2.27
Host machine cpu family: ppc
Host machine cpu: ppc
Did not find pkg-config by name 'pkg-config'
Found Pkg-config: NO
Run-time dependency zlib found: NO (tried pkgconfig)
Run-time dependency libkmod found: NO (tried pkgconfig)
Run-time dependency libudev found: NO (tried pkgconfig)
Has header "resolv.h" : YES
Library resolv found: NO
Checking for function "res_query" with dependency -lresolv: YES
Configuring config.h using configuration
Checking for function "getopt" : YES
Header <unistd.h> has symbol "getopt" : YES
Build targets in project: 10

Found ninja-1.10.2 at /usr/bin/ninja
[27/39] Compiling C object 'pci@sta/lib_sysfs.c.o'
FAILED: pci@sta/lib_sysfs.c.o
gcc -Ipci@sta -I. -I.. -fvisibility=hidden -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -O2 -g -Wno-parentheses -Wstrict-prototypes -Wmissing-prototypes -fPIC -MD -MQ 'pci@sta/lib_sysfs.c.o' -MF 'pci@sta/lib_sysfs.c.o.d' -o 'pci@sta/lib_sysfs.c.o' -c ../lib/sysfs.c
../lib/sysfs.c: In function 'sysfs_read_vpd':
../lib/sysfs.c:569:43: error: unused parameter 'd' [-Werror=unused-parameter]
 static int sysfs_read_vpd(struct pci_dev *d, int pos, byte *buf, int len)
                                           ^
../lib/sysfs.c:569:50: error: unused parameter 'pos' [-Werror=unused-parameter]
 static int sysfs_read_vpd(struct pci_dev *d, int pos, byte *buf, int len)
                                                  ^~~
../lib/sysfs.c:569:61: error: unused parameter 'buf' [-Werror=unused-parameter]
 static int sysfs_read_vpd(struct pci_dev *d, int pos, byte *buf, int len)
                                                             ^~~
../lib/sysfs.c:569:70: error: unused parameter 'len' [-Werror=unused-parameter]
 static int sysfs_read_vpd(struct pci_dev *d, int pos, byte *buf, int len)
                                                                      ^~~
cc1: all warnings being treated as errors
[29/39] Compiling C object 'pci@sta/lib_names.c.o'
ninja: build stopped: subcommand failed.
```

But compiles fine via make.

I have opened PR #158

gollux:
Sorry for too little feedback from me, but I am currently travelling. I will look at this when I return home, thanks for experimenting with it!

Anyway, I don't think that treating warnings as errors is appropriate.

eli-schwartz:
Note that it's a builtin meson option (defaulting to off), so even when removing it from meson.build it's extremely easy to use `meson setup builddir -Dwerror=true` in CI if you'd like to make warnings more noticeable in CI.

Meson actually added this builtin option because we (the Meson developers) know that many people insist on using -Werror in their project files, and although we don't really agree with this decision, we want to make sure it's always trivial for end users to disable -Werror via a single consistent configure time option shared by all meson projects, without needing to sed out build files. So as upstream developers of the build system itself we strongly encourage projects that *do* want to use -Werror by default to do it via the built-in option, but as downstream users of software which is often built by meson we'd kind of be happy if you didn't. :D

rossburton:
I just rewrote our packaging of pciutils and spent a frustrating hour identifying what variables can be set to control the build parameters, so I'm very interested in seeing Meson support merge because no two hand-coded Makefiles are the same, and that means more work for everyone involved.

PR #162, opened 2023-10-24
Add utility for Lane Margining

bombanya:
PCIe Gen 4 spec introduced new extended capability mandatory for all ports - Lane Margining at the Receiver. 
This new feature allows to get an approximation of the Link eye margin diagram by four points.
This information shows how resistant the Link is to external influences and can be useful 
for link debugging and presets tuning. Previously, this information was only available using a hardware debugger. 

PR consists of two parts:
- Patch for `lspci` to add Margining registers reading. There is not much information available without issuing any margining commands, but this info is useful anyway;
- New `pcilmr` utility.

Margining capability assumes the exchange of commands with the device, so its support was implemented as a separate utility `pcilmr`. 
The `pcilmr` allows you to test Links and supports all the features provided by the Margining capability.
The utility is written using a `pcilib`, it is divided into a library part and a main function with 
arguments parsing in the `pciutils` root dir.
Man page is provided as well. 

Utility compilation and man page preparation are integrated into the pciutils `Makefile`, so run 
`make` is enough to start testing the utility (Gen 4/5 device is required though).
Utility was written with Linux in mind and was tested only on this OS.

Example utility results on different systems you can see in [gist](https://gist.github.com/bombanya/f2b15263712757ffba1a11eea011c419).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ LIBNAME=libpci
 
 -include lib/config.mk
 
-PCIINC=lib/config.h lib/header.h lib/pci.h lib/types.h lib/sysdep.h
+PCIINC=lib/config.h lib/header.h lib/pci.h lib/types.h lib/sysdep.h lib/bitops.h
 PCIINC_INS=lib/config.h lib/header.h lib/pci.h lib/types.h
 
 export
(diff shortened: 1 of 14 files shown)


gollux:
Applied version of the patches received by e-mail.

PR #165, opened 2023-11-14
Fix memory leak when fill flags has PCI_FILL_PARENT.

yeahnye:
realpath(xx, NULL) function cause memory alloc, but forget to free them.


gollux:
This is indeed a bug. A fix is welcome, but please follow the indent style of the file you are editing.

yeahnye:
> This is indeed a bug. A fix is welcome, but please follow the indent style of the file you are editing.

of course, it has been done.

gollux:
There were still some whitespace issues (tabs vs. spaces, spaces at the end of line), but I fixed them myself and merged your patch. Thanks!

yeahnye:
Oh, yes, many thanks.

PR #166, opened 2023-12-29
i386-ports: Add support for OpenBSD

DigitalDJ:
This commit adds support for `i386-ports` in OpenBSD. I added this primarily for use with `flashrom`, allowing many more chipsets to be supported.

On OpenBSD, it looks like the arguments to `in` / `out` are reversed. I'm open to suggestions for a better way to handle this, optimally, without touching `lib/i386-ports.c`. I don't write in C often, so this was the quickest way to solve the issue.

### DIFF
--- a/README
+++ b/README
@@ -20,7 +20,7 @@ In runs on the following systems:
 	Linux		(via /sys/bus/pci, /proc/bus/pci or i386 ports)
 	FreeBSD		(via /dev/pci)
 	NetBSD		(via libpci)
-	OpenBSD		(via /dev/pci)
+	OpenBSD		(via /dev/pci or i386 ports)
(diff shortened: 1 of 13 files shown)


DigitalDJ:
After some correspondence with @pali, we have decided to wait until he has refactored `i386-ports` so that the redefinition of `in` / `out` is more elegant. I will re-base against master once the changes are commited.

DigitalDJ:
I've re-based upon a patch provided by @pali Feel free to cherry pick `cc4c248` after you have commited the refactor to i386-ports.

Thank you!

DigitalDJ:
The author and date for the refactoring of i386-io-ports should now be correct.

PR #181, opened 2024-04-26
ls-ecaps: extend decode support for more fields for AER CE and UE status

axiqia:
Extend decode support for more fields for AER CE and UE status prior to PCIe r6.0.


gollux:
Thanks!

PR #184, opened 2024-05-22
dump: allow 6-digit domains

ksztyber:
The SPDK VMD driver assigns domains for the devices behind a VMD by concatenating bus/device/function of the VMD, each on a separate byte. For instance, a device behind a VMD with an address of 5d:05.5 would be assigned domain 5d0505.


gollux:
Thanks, I will pull this.

PR #196, opened 2025-01-14
Add DirectIO library for config space

startergo:


gollux:
I would like to see some rationale for adding this back-end. Poking I/O ports directly from user space is generally a very bad idea and we use it only on platforms where there is no better solution. What is the advantage of this over the win32-kldbg back-end?

PR #197, opened 2025-02-06
lspci: Added decodes for Physical Layer 16/32/64 extended capabilities registers

tristan-tdy:
Added register field decodes in lspci verbose output for the following extended capability registers:
- Physical Layer 16 GT/s
- Physical Layer 32 GT/s
- Physical Layer 64 GT/s

Also added missing register names to setpci for these registers.


gollux:
Could you please add a test case to `tests/`?

tristan-tdy:
@gollux New PR for requested test case #206

PR #199, opened 2025-02-10
Add `extern "C"` for use with C++

SoapGentoo:
* Without this, we get linking issues:

    simple_lspci.cpp:(.text+0x20): undefined reference to `pci_alloc()'

### DIFF
--- a/lib/pci.h
+++ b/lib/pci.h
@@ -18,6 +18,10 @@
 #include "header.h"
 #include "types.h"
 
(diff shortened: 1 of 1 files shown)


gollux:
Fixed in a slightly different way.

JohnAZoidberg:
For reference, this is the change: https://github.com/pciutils/pciutils/commit/0fb1f314092233b6dc03e4dd3ce42e8c7b1f6304

PR #201, opened 2025-04-17
Port to RT-Thread Smart DM PCI

GuEe-GUI:
RT-Thread support PCI (Only Device Driver Mode) in version 5.2.

We make a port for RT-Smart (have user mode version in RT-Thread), and add the PCI/PCIe access in `/proc/pci` for user like linux but some different:

https://github.com/RT-Thread/rt-thread/pull/10196

### DIFF
--- a/README
+++ b/README
@@ -33,6 +33,7 @@ In runs on the following systems:
 	DOS/DJGPP	(via i386 ports)
 	SylixOS		(via /proc/pci)
 	AmigaOS on PPC	(via Expansion library)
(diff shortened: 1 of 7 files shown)


GuEe-GUI:
Maybe you don't want to check the RT-Thread content, so I’m directly providing the running results here:

lspci

gollux (changes requested):
Thanks for your contribution!

Could you please update the documentation in `pcilib.man`, so that it includes your back-end and its configurable parameters?

PR #202, opened 2025-04-29
ls-caps: add FPB capability

bwlodarcz:
Flattening Portal Bridge (FPB) is mechanism which changes the way the routing ID's
and memory resources are assigned. Bus Numbers are flattened inside switches
and downstream ports which allows to allocate resources in non-continuous ranges.
It's optional by standard and aims in improvement of scalability and enhance hot-add
ability in PCIe. 
Sec 6.21 in PCIe ver. 6+ specification.

### DIFF
--- a/ls-caps.c
+++ b/ls-caps.c
@@ -1792,6 +1792,321 @@ static void cap_ea(struct device *d, int where, int cap)
   }
 }
 
+static void cap_fpb(struct device *d, int where, int type)
+{
+  if (type != PCI_EXP_TYPE_DOWNSTREAM && type != PCI_EXP_TYPE_UPSTREAM)
+    return;
+  if (!config_fetch(d, where + 4, 28))
+    return;
+  u32 cap_register = get_conf_long(d, where + 4);
+  u32 rid_vec_control_1_reg = get_conf_long(d, where + 8);
+  u32 rid_vec_control_2_reg = get_conf_long(d, where + 12);
+  u32 mem_low_control_reg = get_conf_long(d, where + 16);
+  u32 mem_high_control_1_reg = get_conf_long(d, where + 20);
+  u32 mem_high_control_2_reg = get_conf_long(d, where + 24);
+  u32 vec_access_control_reg = get_conf_long(d, where + 28);
+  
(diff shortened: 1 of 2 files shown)


bwlodarcz:
@gollux I want to contribute this cap when it'll be ready. 
I have two questions:
a) Hot testing works in pciutils. I see that there is a test folder but can't find any info how to trigger them (nothing in make) and how to write my own.
b) Is there any guide/hints how output should look like.

ls-caps.c
```diff
@@ -1792,6 +1792,321 @@ static void cap_ea(struct device *d, int where, int cap)
...
+  int mem_high_decode_supported = cap_register & 4;
+  int rid_vector_size_supported = BITS(cap_register, 8, 3);
+  int mem_low_vector_size_supported = BITS(cap_register, 16, 3);
+  int mem_high_vector_size_supported = BITS(cap_register, 24, 3);
+  puts("Flattening Portal Bridge");
+  if (type == PCI_EXP_TYPE_UPSTREAM) {
+    int num_sec_devices = BITS(cap_register, 3, 5);
+    printf("\t\tNum Sec Dev: %d\n", num_sec_devices + 1);
+  }
+  const char *vector_access_offset;
```
gollux, line 1820: Please use blank lines to split the code to logical parts.

ls-caps.c
```diff
@@ -1792,6 +1792,321 @@ static void cap_ea(struct device *d, int where, int cap)
...
+  switch(BITS(vec_access_control_reg, 0, 8)) {
+  case 0:
```
gollux, line 1822: Perhaps we should index a table instead?

ls-caps.c
```diff
@@ -1792,6 +1792,321 @@ static void cap_ea(struct device *d, int where, int cap)
...
+    vector_access_offset = "6:0 (7 unused)";
+    break;
+  case 5:
+    vector_access_offset = "7:0";
+    break;
+  default:
+    vector_access_offset = "<reserved>";
+  }
+  const char *vector_access_select;
+  switch(BITS(vec_access_control_reg, 14, 2)) {
```
gollux, line 1844: The same here.

PR #203, opened 2025-04-29
Fix lack of exposure of pci_init for LIBPCI_3.0 on pciutils-3.13.0

listout:
The changes included in commit 25471140536014e21c7402f6783bf882ed02aec2 were not sufficient to maintain ABI compatibility with older versions of libpci.so.3

The idea comes directly from commit 25471140536014, and the patch from Ian Stakenvicius[0].

[0]: https://www.spinics.net/lists/linux-pci/msg51269.html


listout:
We caught this after one of your customers at SUSE reported that biosdevname stopped working after pciutils update to 3.5.6. As I mentioned I crafted this patch taking ideas from previous patch/commit and would love some feedback. Since we like to keep backwards compatibility.

Thank you!

listout:
@peti

peti:
I believe the PR title and commit description should both say LIBPCI_3.2 instead of LIBPCI_3.0.

listout:
Fixed, thanks!

peti:
The 3.2 function is an alias for pci_init_v35().

gollux:
In pciutils-3.2.0, `pci_init()` is versioned as `LIBPCI_3.0`, because the function was not modified since pciutils-3.0.

peti:
I believe the LIBPCI_3.2 namespace is generated by libtool since that is the public API version of the library? I'm not entirely sure though. All I can say is that we had binaries linked to libpci 3.2 that failed to resolve the init function after the system updated to libpci version 3.5.x.

gollux:
@listout Could you please find out how was the binary produced?

PR #206, opened 2025-06-09
lspci: Add test case for Physical Layer 16 GT/s and 32 GT/s extended capability registers

tristan-tdy:
Added test case requested from #197


gollux:
Thanks!

PR #207, opened 2025-06-18
Use snprintf instead of sprintf to prevent buffer overruns

szsam:
In bitops.h, update the TABLE macro to call snprintf(buf, sizeof(buf), ...) rather than unbounded sprintf, ensuring that out-of-range indices produce a bounded "??%d" string.

In setpci.c, change the device slot formatting from sprintf(slot, ...) to snprintf(slot, sizeof(slot), ...), capping output to the 16-byte buffer and avoiding overflow when printing PCI domain, bus, dev, and func values.


gollux:
We used to support platforms without `snprintf`, but it's no longer the case. Thanks!

PR #218, opened 2026-01-25
lspci: add test with capabilities 32, 34, and 36

alexmartens-asteralabs:
- h0032: Flit logging
- h0034: Flit error injection
- h0036: MMIO register block locator

---

This is a follow up to the decoders added #209, #212, and #213.


gollux:
Thanks!

PR #226, opened 2026-05-06
pcilmr: Add PCIe Gen 6 (64 GT/s) support

astera-kbohlen:
- Extend speed-indexed arrays in lmr.h to support link_speed=6
- Allow Gen 6 link speed in margin_verify_link()
- Fix NULL dereference crash in margin_fill_link() when either port lacks LMR capability
- Fix scan_links() and find_ready_links() to check up port has LMR capability before reporting link as available
- Fix GT/s display formula in margin_log_link() for Gen 6
- Update user-facing strings to include 64 GT/s


gollux:
Thanks, merged.

PR #229, opened 2026-06-11
fix: add bounds checking to parse_csv_arg to prevent buffer overflow

xuammu:
Add max_vals parameter to parse_csv_arg() and enforce it in the parsing loop to avoid writing past the end of the lanes/recvs arrays when CSV input contains too many values.

### DIFF
--- a/lmr/margin_args.c
+++ b/lmr/margin_args.c
@@ -60,16 +60,17 @@ dev_for_filter(struct pci_access *pacc, char *filter)
 }
 
 static u8
-parse_csv_arg(char *arg, u8 *vals)
+parse_csv_arg(char *arg, u8 *vals, u8 max_vals)
 {
   u8 cnt = 0;
   char *token = strtok(arg, ",");
-  while (token)
-    {
-      vals[cnt] = atoi(token);
-      cnt++;
-      token = strtok(NULL, ",");
-    }
+  while (token && cnt < max_vals) {
+    vals[cnt] = atoi(token);
+    cnt++;
+    token = strtok(NULL, ",");
+  }
+  if (token)
+    die("Too many values in CSV argument. Maximum is %u.\n", max_vals);
   return cnt;
 }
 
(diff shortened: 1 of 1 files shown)


lmr/margin_args.c
```diff
@@ -59,18 +59,17 @@ dev_for_filter(struct pci_access *pacc, char *filter)
   die("No such PCI device: %s or you don't have enough privileges.\n", filter);
 }
 
-static u8
-parse_csv_arg(char *arg, u8 *vals)
+static u8
```
gollux: Please do not add trailing spaces.

lmr/margin_args.c
```diff
@@ -59,18 +59,17 @@ dev_for_filter(struct pci_access *pacc, char *filter)
...
-  u8 cnt = 0;
-  char *token = strtok(arg, ",");
-  while (token)
-    {
-      vals[cnt] = atoi(token);
-      cnt++;
-      token = strtok(NULL, ",");
+    u8 cnt = 0;
+    char *token = strtok(arg, ",");
+    while (token && cnt < max_vals) {
```
gollux: Please follow indentation style of the rest of the file.

lmr/margin_args.c
```diff
@@ -59,18 +59,17 @@ dev_for_filter(struct pci_access *pacc, char *filter)
...
-      cnt++;
-      token = strtok(NULL, ",");
+    u8 cnt = 0;
+    char *token = strtok(arg, ",");
+    while (token && cnt < max_vals) {   
+        vals[cnt] = atoi(token);
+        cnt++;
+        token = strtok(NULL, ",");
     }
-  return cnt;
```
gollux, line 73: If the limit is exceeded, you should report an error instead of silently discarding extra data.

lmr/margin_args.c
```diff
@@ -132,10 +131,10 @@ parse_dev_args(int argc, char **argv, struct margin_link_args *args, u8 link_spe
             args->parallel_lanes = atoi(optarg);
             break;
           case 'l':
-            args->lanes_n = parse_csv_arg(optarg, args->lanes);
+            args->lanes_n = parse_csv_arg(optarg, args->lanes, 32);
```
gollux: This (and the "8" below) should not be hard-coded. Use `sizeof(args->lanes) / sizeof(*args->lanes)`.

gollux (changes requested):
Thanks for reporting the problem. However, your fix needs cleaning up before it can be merged.

xuammu:
all fixed.

PR #230, opened 2026-07-03
pcilmr: update PCIe 6.0 margin handling

astera-kbohlen:
This updates pcilmr's PCIe 6.0 margin offset handling.

Changes:
- use PAM4-aware max voltage offset scaling for Gen6+
- use PAM4-aware max timing offset fallback for Gen6+
- update warning text for unreported timing/voltage offsets
- update Gen6-related UI / grading reference values in lmr.h

Background:
For PCIe 6.0 PAM4 operation, Max Voltage Offset is defined as a percentage of one third volt rather than one volt as in NRZ. The previous code used NRZ-style scaling unconditionally, which overstated Gen6 voltage margins.

### DIFF
--- a/lmr/lmr.h
+++ b/lmr/lmr.h
@@ -18,7 +18,7 @@
 enum margin_hw { MARGIN_HW_DEFAULT, MARGIN_ICE_LAKE_RC };
 
 // in ps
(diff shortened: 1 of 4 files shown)


gollux:
Thanks, merging.

PR #235, opened 2026-09-19
Add PCI_FILL_SIZES support in generic

sthibaul:


gollux:
Sorry, but this is not reasonable. `pci_fill_info()` must _never_ modify configuration space.

sthibaul:
Can `init_dev` do it ahead of the `fill_info` call?

gollux:
No. The `libpci` is not supposed to write anything to the config space unless you ask for it explicitly. What if any other program (e.g., the OS) is using the device at the same time?

sthibaul:
> What if any other program (e.g., the OS) is using the device at the same time?

Ok, the use case is MiniOS where there is no other program :) Maybe I can just stuff this into a `minios.c`-specific `fill_info` or `init_dev` method?

PR #236, opened 2026-09-19
Add Xen MiniOS support

sthibaul:


lib/sysdep.h
```diff
@@ -101,6 +101,10 @@ typedef u16 word;
   #include <machine/endian.h>
 #endif
 
+#ifdef PCI_OS_HURD
```
gollux: Is this related to the rest of the commit?

lib/init.c
```diff
@@ -168,6 +168,11 @@ static struct pci_methods *pci_methods[PCI_ACCESS_MAX] = {
 #else
   NULL,
 #endif
+#ifdef PCI_OS_MINIOS
```
gollux, line 171: This does not seem to be defined anywhere.

lib/init.c
```diff
@@ -187,6 +192,7 @@ static int probe_sequence[] = {
   PCI_ACCESS_WIN32_SYSDBG,
   PCI_ACCESS_AOS_EXPANSION,
   PCI_ACCESS_RT_THREAD_SMART_DM,
+  PCI_ACCESS_MINIOS,
```
gollux, line 195: And neither does this.

gollux (changes requested):
Generally, I welcome the new access method. But it couple of fixes first - see the review comments. Also, you should add the method to the pcilib man page.

lib/init.c
sthibaul: ah, missing hook indeed

lib/init.c
sthibaul: Yes, that is sort of expected. In the minios case we cross-build the library, but using the host compiler with just different header/library paths, so `lib/configure` can not really detect it by itself, so it's the cross-compilation script that sets `PCI_OS_MINIOS=1` by hand.

lib/sysdep.h
sthibaul: actually yes, I just mixed up some hook with the Hurd part which I happened to get a look at at the same time.

lib/init.c
gollux: Then you should document the build process (e.g., in the README).

sthibaul: ok, done so

